PR #6, opened 2025-01-17
Draft updates 2

Vectorized:
- ERC20 and native token spend permissions.

Todo:
- Write tests.
- EntryPoint create2 address.
- Ability to let delegation approve more than one EntryPoint (tied to current delegation codehash). This allows for the delegation to possibly switch to a more performant, but abi equivalent EntryPoint in the future.

Thoughts:
- The EntryPoint is quite a complex beast, due to it's need for gas introspection and performance, just like the 4337 EntryPoint. I think in the end state, a good EntryPoint will be written in very optimized inline-assembly. Also, the main bottleneck in Base's gas costs by far is L2 gas.

### DIFF
--- a/src/Delegation.sol
+++ b/src/Delegation.sol
@@ -29,7 +29,7 @@ contract Delegation is EIP712, GuardedExecutor {
     /// @dev The type of key.
     enum KeyType {
         P256,
-        WebAuthnP256, 
+        WebAuthnP256,
         Secp256k1
     }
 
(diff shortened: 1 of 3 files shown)


gakonst (commented):
WDYT about the AA folks exposing the validation logic in a separate function? Is it worth it?

I think there is a benefit because it'd let us build a Reth executor which prefetches session keys / verifies sigs in parallel, which could improve performance

I know it introduces the pre verification gas stuff which IMO should not be exposed to devs so I'd like to avoid that if we thought doing the validation / execution separation was worth it.

gakonst (approved):
concept ACK & supportive of merging, we can iterate on it. i'd like us to be thoughtful about how the key is laid out in storage, so that proving storage proofs about a session key + its policies is easy/optimized

PR #21, opened 2025-02-05
fix: prevent division by zero in _brutalizeMemory()

NikolaiKryshnev:
Added a safety check to ensure `cSize` is never zero in `_brutalizeMemory()`.  
This prevents potential division by zero errors when calculating `mod(mload(0x10), cSize)`,  
which could cause unexpected behavior in memory brutalization tests.

Changes:
- Added `if eq(cSize, 0) { cSize := 1 }` to `_brutalizeMemory()`.


Vectorized:
Opcode level division by zero gives zero, which is within expected range.

jenpaff:
> Opcode level division by zero gives zero, which is within expected range.

After @Vectorized feedback I believe we can close this one, feel free to open a new PR including a test if incorrect.

PR #29, opened 2025-02-13
♻️ Added Test for EntryPoint & Fix Issues

atarpara:
**Summary:**  
- Added test cases for Entrypoint :  
  - [x] Sponsored transactions  
  - [x] Batch calls  
  - [x] Relayer receiving sufficient funds for gas payment  
  - [x] Edge cases related to gas 
  - [x] Refund calculation  
  
 Fix Issue: 
 1. Fixed `Entrypoint._verify()` signature length  
 2. Fix wrong error message


onbjerg (commented):
this is pretty good although i'd love if we also had tests where the signature isnt just rsv since that only works for the eoa root key as far as i understand. i.e. if we had tests for a secp256k1 key that doesnt recover to `address(eoa)`. same with p256 and p256webauthn (although i guess the last one is pretty hard to achieve in solidity?)

atarpara:
> this is pretty good although i'd love if we also had tests where the signature isnt just rsv since that only works for the eoa root key as far as i understand. i.e. if we had tests for a secp256k1 key that doesnt recover to `address(eoa)`. same with p256 and p256webauthn (although i guess the last one is pretty hard to achieve in solidity?)

Added P256 & secp256k1 separate testcases. While p256webauthn is hard to test in solidity working on it.

Vectorized (approved):

Vectorized:
I think let's merge this to main for now. We can revert if something goes wrong.

After this is merged, I'll make the PR for the new upgradable proxy.

PR #30, opened 2025-02-17
♻️ Refactor EntryPoint nonces

atarpara:
Rationale:

Right now, the EntryPoint nonce system is in the Delegation. This means that when a transaction fails, the ERC20 payment is actually made, but the nonce isn't consumed. This is dangerous, since the signed userop can be abused to drain the ERC20 of the eoa.

This PR does 3 things.

1. Move the EntryPoint nonce system to the EntryPoint itself. This means if a UserOp fails, the nonce will still be consumed.

2. Change the nonce style to be a 4337-style 2D nonce [more info](https://hackmd.io/@atarpara/BkU-NNz5yl). 
 - Previously, we can invalidate all the unused signatures by simply incrementing `nonceSalt`. This 2D nonce system does not have `nonceSalt`. This means to invalidate all the unused signatures, one must take note of all the `keySeq` (i.e. the upper 192 bits that have been signed), and invalidate them one-by-one.
 - For out-of-order nonces, just make the `keySeq` pseudorandom. I have yet to determine a way to mass invalidate nonces across `keySeq`s.
3. On UserOp failure, record the `err` for the nonce in storage.
```
storageSlot = keccak256(
    abi.encode(
        (userOp.nonce >> 64),
        keccak256(
            abi.encode(
                userOp.eoa,
                uint72(bytes9(keccak256("PORTO_ENTRY_POINT_STORAGE"))) + 1
            )
        )
    )
);
```

- [x] Added sequential nonce increment for each eoa
- [x] Remove `_useNonce` from delegation.sol for `entryPoint`
- [x] Fix Entrypoint Test

### DIFF
--- a/src/Delegation.sol
+++ b/src/Delegation.sol
@@ -64,10 +64,8 @@ contract Delegation is EIP712, GuardedExecutor {
     struct DelegationStorage {
         /// @dev The label.
         LibBytes.BytesStorage label;
-        /// @dev Bitmap of invalidated nonces. Set bit means invalidated.
-        LibBitmap.Bitmap invalidatedNonces;
-        /// @dev The current nonce salt.
-        uint256 nonceSalt;
+        /// @dev Account nonce.
+        mapping(uint192 => uint64) nonce;
         /// @dev Set of key hashes for onchain enumeration of authorized keys.
         EnumerableSetLib.Bytes32Set keyHashes;
         /// @dev Mapping of key hash to the key in encoded form.
@@ -147,12 +145,6 @@ contract Delegation is EIP712, GuardedExecutor {
     /// @dev The key with a corresponding `keyHash` has been revoked.
     event Revoked(bytes32 indexed keyHash);
 
-    /// @dev The `nonce` have been invalidated.
-    event NonceInvalidated(uint256 nonce);
-
-    /// @dev The nonce salt has been incremented to `newNonceSalt`.
-    event NonceSaltIncremented(uint256 newNonceSalt);
-
     /// @dev The `checker` has been authorized to use `isValidSignature` for `keyHash`.
     event SignatureCheckerApprovalSet(
         bytes32 indexed keyHash, address indexed checker, bool isApproved
@@ -167,7 +159,7 @@ contract Delegation is EIP712, GuardedExecutor {
 
     /// @dev For EIP712 signature digest calculation for the `execute` function.
     bytes32 public constant EXECUTE_TYPEHASH = keccak256(
-        "Execute(bool multichain,Call[] calls,uint256 nonce,uint256 nonceSalt)Call(address target,uint256 value,bytes data)"
+        "Execute(bool multichain,Call[] calls,uint256 nonce)Call(address target,uint256 value,bytes data)"
     );
 
     /// @dev For EIP712 signature digest calculation for the `execute` function.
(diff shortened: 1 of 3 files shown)


src/Delegation.sol
```diff
@@ -564,7 +564,6 @@ contract Delegation is EIP712, GuardedExecutor {
         // Entry point workflow.
         if (msg.sender == ENTRY_POINT) {
             if (opData.length < 0x40) revert OpDataTooShort();
```
Vectorized, line 566: `opData.length < 0x20`, since we don't need to pass the nonce from the EntryPoint to the Delegation anymore.

src/EntryPoint.sol
```diff
@@ -133,6 +136,7 @@ contract EntryPoint is EIP712, Ownable, CallContextChecker, ReentrancyGuardTrans
     /// @dev Holds the storage.
     struct EntryPointStorage {
         LibBitmap.Bitmap filledOrderIds;
+        mapping(address => mapping(uint192 => uint64)) accountsNonce;
```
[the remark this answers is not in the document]
Vectorized: TODO: write a short paragraph on how to use ERC-4337 style 2D nonces.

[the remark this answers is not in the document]
Vectorized: I think `accountsNonce` should be placed at position 0. 

I am not entirely sure if we actually will go with cross-chain filling in the final version.

src/EntryPoint.sol
```diff
@@ -466,32 +482,22 @@ contract EntryPoint is EIP712, Ownable, CallContextChecker, ReentrancyGuardTrans
                 );
             }
         }
+        bool isMultichain =
+            (u.nonce >> 224) == uint256(0xb014aac5b0d6328e584ecc632bc05bad2f720d727ce21382);
```
Vectorized, line 486: I think we should only use the upper 4 bytes of the upper 192 bits to denote multichain.

Otherwise, someone can simply invalidate the entire multichain nonce dimension by setting the nonce to `2 ** 64 - 1`, rendering it useless.

Vectorized (changes requested):
Good work.

Let's also refactor the delegation to use 4337-style 2D nonces.

Vectorized:
@atarpara Add the part on recording the `err` in storage if the UserOp fails.

Vectorized:
@atarpara I am thinking of reserving 32 bits in the `keySeq` to enable one-dot mass invalidation of nonces across all nonces.

src/Delegation.sol
Vectorized: Do remember to update the EntryPoint accordingly to remove the `nonce` from the `abi.encode(keyHash, nonce)`.

src/EntryPoint.sol
atarpara: I don't think 4 bytes are sufficient for `multichain` operation because it will accidentally or intentionally collide with other keys or be exploited by a bad actor using finding collision with other word.

src/Delegation.sol
atarpara: > Do remember to update the EntryPoint accordingly to remove the `nonce` from the `abi.encode(keyHash, nonce)`.

where it is i didn't get.

atarpara:
> @atarpara Add the part on recording the `err` in storage if the UserOp fails.

Added.

atarpara:
@Vectorized
If `_pay()` fails, the nonce will not be consumed because it will add an extra 5000 gas overhead to the relayer without getting paid.

atarpara:
> @atarpara I am thinking of reserving 32 bits in the `keySeq` to enable one-shot mass invalidation of nonces across all nonces.

can you explain bit more.

Vectorized:
@atarpara Made the changes.

Vectorized:
Ok, let's merge this for now. If anything goes wrong, we still can revert.

PR #36, opened 2025-02-20
Events for nonces

Vectorized:


Vectorized:
Thinking:

Would having just 1 event be easier to work with? 

Do we need err?

Vectorized:
Merging this first.

PR #55, opened 2025-02-25
Verification agnostic gas estimation

Vectorized:
So that we don't need to go thru all those signature guessing stuff.

This PR still simulates the gas burn from the signature validation, but it avoids reverting if the signature is invalid. This allows the relay to simulate the guarded executor overhead for the particular `keyHash`, without needing a proper signature from that `keyHash`.


onbjerg (commented):
lets document on `simulateExecute2` how the spoofing works for `GuardedExecutor`

Vectorized:
@onbjerg Added to the Natspec.

PR #56, opened 2025-02-26
v0.0.1

Vectorized:
Based off #54 #55.

Todo:
- [ ] More GuardedExecutor tests


Vectorized:
Adding a version from now on so that changelogs can be auto generated via tags.

PR #57, opened 2025-02-26
PREP workflow and fix payment draining vector

Vectorized:
https://blog.biconomy.io/prep-deep-dive/#nicks%E2%80%99-method

Adds the PREP workflow, which relies on a new method `initializePREP` on the Delegation, and a new `initData` field in the UserOp.

____

Update: This PR also fixes an attack vector where the user can be drained. 

There’s a chicken and egg problem. Validation of the UserOp is needed to ensure that `_pay` deducts correctly, but `_verify` costs variable amount of gas, which can be griefed.

Suggestion: simulate off-chain to ensure that the UserOp passes.
If off-chain simulation passes, but fails on-chain due to validation error, suspect griefing and ban/throttle the user.

For extra safety, the off-chain simulation can even check the 7702 delegation and implementation slot. This further reduces the griefing chance. Even though the user can quickly swap out their implementation or delegation in the small window between off-chain simulation and txn execution, they will need to pay gas for the swapping txn. The amount of gas required to swap out the implementation is greater than the amount of gas the relayer risks losing, so this is still acceptable.

Todo:
- [x] Write some very basic tests.


Vectorized:
I think i'll merge this first, then #60 (cuz that's breaking).

PR #63, opened 2025-03-06
More tests and tidy

Vectorized:
- Refactor and add more tests.
- Refactor nonce logic into a LibNonce, so we can avoid having to duplicate testing for Delegation.
- Make `canExecute` mapping enumerable.
- Fixed a bug where it is not possible to add `ANY_KEYHASH` due to `getKey` reverting for missing key.


gakonst (approved):
There's a lot to go over here. Merging it in so we can be unblocked and we'll do a synchronous review over them.

PR #65, opened 2025-03-07
Fix removeSpendLimit and add test

Vectorized:
Continued from #64.

### DIFF
--- a/src/GuardedExecutor.sol
+++ b/src/GuardedExecutor.sol
@@ -324,12 +324,13 @@ contract GuardedExecutor is ERC7821 {
         checkKeyHashIsNonZero(keyHash)
     {
         SpendStorage storage spends = _getGuardedExecutorStorage().spends[keyHash];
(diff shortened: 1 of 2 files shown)


src/GuardedExecutor.sol
```diff
@@ -324,12 +324,14 @@ contract GuardedExecutor is ERC7821 {
         checkKeyHashIsNonZero(keyHash)
     {
         SpendStorage storage spends = _getGuardedExecutorStorage().spends[keyHash];
-        spends.tokens.remove(token);
 
         TokenSpendStorage storage tokenSpends = spends.spends[token];
-        tokenSpends.periods.remove(uint8(period));
+        if (tokenSpends.periods.remove(uint8(period))) {
+            if (tokenSpends.periods.length() == uint256(0)) spends.tokens.remove(token);
+        }
```
gakonst, line 332: yep good catch

gakonst (approved):
good find

PR #66, opened 2025-03-07
Refactor deployment, EntryPoint, Delegation

Vectorized:
- Remove fallback from EntryPoint. Use mined functions instead.
- Refactor deployment scripts.
- Use latest Solady which has a EIP7702Proxy that has the `defaultImplementation` in an immutable if the proxy has no `initialAdmin`.
- Make `ENTRY_POINT` immutable on Delegation, to be set via constructor.


src/EntryPoint.sol
```diff
@@ -674,80 +762,8 @@ contract EntryPoint is EIP712, Ownable, CallContextChecker, ReentrancyGuardTrans
...
 
-    ////////////////////////////////////////////////////////////////////////
-    // Fallback
-    ////////////////////////////////////////////////////////////////////////
-
     receive() external payable virtual {}
 
-    /// @dev Use the fallback function to implement gas limited verification and execution.
-    /// Helps avoid unnecessary calldata decoding.
-    fallback() external payable virtual {
```
gakonst, line 685: thank god

src/EntryPoint.sol
```diff
@@ -443,6 +453,93 @@ contract EntryPoint is EIP712, Ownable, CallContextChecker, ReentrancyGuardTrans
...
+        require(msg.sender == address(this));
+
+        UserOp calldata u;
+        bytes32 keyHash;
+        assembly ("memory-safe") {
+            u := add(0x24, calldataload(0x24))
+            keyHash := calldataload(0x04)
+        }
+        _execute(u, keyHash, false);
+    }
```
gakonst, line 541: Do we need to have a separate function for `selfCallExecute` or can you just inline it above?

src/EntryPoint.sol
```diff
@@ -443,6 +453,93 @@ contract EntryPoint is EIP712, Ownable, CallContextChecker, ReentrancyGuardTrans
...
+        require(msg.sender == address(this));
+
+        UserOp calldata u;
+        bytes32 isGasSimulation;
+        assembly ("memory-safe") {
+            u := add(0x24, calldataload(0x24))
+            isGasSimulation := calldataload(0x04)
+        }
+        // Verify the nonce, early reverting to save gas.
+        (LibStorage.Ref storage s, uint256 seq) =
```
gakonst, line 489:
```suggestion
        (LibStorage.Ref storage nonce, uint256 seq) =
```

no more 1-letter variable names!

src/EntryPoint.sol
```diff
@@ -443,6 +453,93 @@ contract EntryPoint is EIP712, Ownable, CallContextChecker, ReentrancyGuardTrans
...
 
+    /// @dev This function is only intended for self-call.
+    /// The name is mined to give a function selector of `0x00000000`, which makes it
+    /// more efficient to call by placing it at the leftmost part of the function dispatch tree.
+    ///
+    /// We perform a gas-limited self-call to this function via `_execute(bytes,uint256)`
+    /// with assembly for the following reasons:
+    /// - Allow recovery from out-of-gas errors.
+    ///   When a transaction is actually mined, an `executionData` payload that takes 100k gas
+    ///   to execute during simulation might require 1M gas to actually execute.
```
gakonst, line 464: what does this mean? when?

script/DeployAll.s.sol
```diff
@@ -2,13 +2,14 @@
...
 
 import {Script} from "forge-std/Script.sol";
-import "../src/Delegation.sol";
+import "../src/DeployAll.sol";
 
-contract DeployDelegationImplementationScript is Script {
+contract DeployAllScript is Script {
     function run() external {
         uint256 deployerPrivateKey = vm.envUint("PRIVATE_KEY");
+        address deployer = vm.createWallet(deployerPrivateKey).addr;
```
gakonst, line 10: it's an anti-pattern to be doing this stuff IMO and you should be using `vm.startBroadcast()` and rely on foundry to inject the right signer here, either via `--private-key` or via `--aws` etc.

gakonst (approved):
good progress.

src/EntryPoint.sol
Vectorized: seqRef

src/EntryPoint.sol
Vectorized: Need try. Superficial gut feel says making it inline is possible.

PR #78, opened 2025-03-18
fix: fill `gUsed` on sponsored executes

joshieDo:
We use paymentAmount=0 for estimations, and this messes up with simulations (and sponsored executes)


gakonst (approved):
LGTM & makes sense,

onbjerg (approved):

PR #82, opened 2025-03-19
feat: add `spendAndExecuteInfos` to `Delegation` contract

joshieDo:
To be used only on `eth_call` to fetch all spend and execute permissions from a set of keys.


mattsse (approved):
this looks fine,

wdyt @klkvr

klkvr (approved):
lgtm

Issue #84, opened 2025-03-20
Add verification to auto deployments


joshieDo:
https://github.com/ithacaxyz/account/pull/83#issuecomment-2740520700

PR #85, opened 2025-03-20
feat: verify contracts in deploy workflow

klkvr:
Blockscout does not accept `prague` EVM version by default. This should be fixed with https://github.com/blockscout/blockscout/pull/12115. For now this PR sets the `evm_version` to `cancun` as `prague` is basically the same thing from solc pov

Closes https://github.com/ithacaxyz/account/issues/84


onbjerg (commented):
smol nit, let's do `--evm-version cancun` for `forge script` instead

PR #87, opened 2025-03-20
fix: use nightly foundry for deploy

klkvr:
We need https://github.com/foundry-rs/foundry/pull/9979 which is not available in `stable` foundry for now, hence verification is silently failing


onbjerg (approved):

onbjerg:
let's note to go back on stable after 9979 is released

PR #89, opened 2025-03-21
PreOps on EntryPoint

Vectorized:
This adds the BoB (batch-of-batches) to the EntryPoint.


src/EntryPoint.sol
```diff
@@ -649,6 +672,57 @@ contract EntryPoint is
...
+                hex"01000000000078210001", // ERC7821 batch execution mode.
+                u.executionData,
+                abi.encode(keyHash) // `opData`.
+            );
+            // This part is slightly different from `selfCallPayVerifyCall537021665`.
+            // It always reverts on failure.
+            assembly ("memory-safe") {
+                mstore(0x00, 0) // Zeroize the return slot.
+                if iszero(call(gas(), eoa, 0, add(0x20, data), mload(data), 0x00, 0x20)) {
+                    // If this is a simulation via `simulateFailed`, bubble up the whole revert.
```
gakonst, line 711: What should the gas limit be here? We cannot forward everything here. My suggestion was that we cap each PreOp to 200K gas (or whatever the most expensive key mgmt operation is?) so the relay can predict how much they can get grieved for in the worst case? The relay can also just have a simple PreOps whitelist to ensure they don't call any unsafe method.

gakonst (approved):
We also need to enforce that PreOps have paymentAmount=0, but can go separately

src/EntryPoint.sol
```diff
@@ -649,6 +673,53 @@ contract EntryPoint is
...
+        uint256 simulationFlags,
+        bytes[] calldata encodedSubUserOps
+    ) internal virtual {
+        for (uint256 i; i < encodedSubUserOps.length; ++i) {
+            UserOp calldata u = _extractUserOp(encodedSubUserOps[i]);
+            if (eoa != u.eoa) revert InvalidSubUserOpEOA();
+
+            if (u.encodedSubUserOps.length != 0) {
+                _handleEncodedSubUserOps(eoa, simulationFlags, u.encodedSubUserOps);
+            }
```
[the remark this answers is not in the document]
Vectorized: It's to allow the function to be as general as possible. 

For example, one can make a PreOp that authorize a single passkey, and another PreOp containing that to add some spend permissions.

src/EntryPoint.sol
Vectorized: We can cap it to 200k gas.

Vectorized:
Payment amounts in PreOps will be totally ignored.

Use case:

Someone signs a UserOp to authorize a single key, which contains all the payment stuff required.

This UserOp can submitted by itself on one chain, and submitted as a PreOp on another chain.

PR #102, opened 2025-04-03
feat: `simulateExecute` no longer reverts

joshieDo:
This enables the usage of `eth_simulateV1` which collects all logs from execution. For that we need a simulation that does not revert.

### DIFF
--- a/src/EntryPoint.sol
+++ b/src/EntryPoint.sol
@@ -126,17 +126,17 @@ contract EntryPoint is
     }
 
     ////////////////////////////////////////////////////////////////////////
-    // UserOp offets
+    // UserOp Offets
     ////////////////////////////////////////////////////////////////////////
 
     /// @dev Offset of `paymentAmount` in the UserOp struct.
-    uint256 internal constant _PAYMENT_AMOUNT_OFFSET = 6 * 0x20;
+    uint256 internal constant _USER_OP_PAYMENT_AMOUNT_OFFSET = 6 * 0x20;
 
     /// @dev Offset of `paymentMaxAmount` in the UserOp struct.
-    uint256 internal constant _PAYMENT_MAX_AMOUNT_OFFSET = 7 * 0x20;
+    uint256 internal constant _USER_OP_PAYMENT_MAX_AMOUNT_OFFSET = 7 * 0x20;
 
     /// @dev Offset of `paymentPerGas` in the UserOp struct.
-    uint256 internal constant _PAYMENT_PER_GAS_OFFSET = 8 * 0x20;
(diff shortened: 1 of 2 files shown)


Vectorized (approved):
lgtm

src/EntryPoint.sol
```diff
@@ -411,8 +461,14 @@ contract EntryPoint is
...
             (gUsed, err) = _execute(encodedUserOp, combinedGasOverride);
         }
-        // Revert with `abi.encodePacked(bytes4(0xffffffff), abi.encode(gUsed, err))`.
         assembly ("memory-safe") {
+            let noRevertCaller := calldataload(0x24)
+            if noRevertCaller {
+                // Require that the balance of the no revert caller is `type(uint256).max`.
+                if add(balance(noRevertCaller), 1) { invalid() }
+                return(0x00, 0x00) //
+            }
```
Vectorized, line 470: Note: `type(uint256).max + 1 == 0` (due to overflow).

test/utils/mocks/MockEntryPoint.sol
```diff
@@ -7,7 +7,20 @@ import {Brutalizer} from "../Brutalizer.sol";
...
+        uTest.paymentPerGas = 0x334455;
+        bytes memory encoded = abi.encode(uTest);
+        assembly ("memory-safe") {
+            let o := add(encoded, 0x20)
+            let u := add(o, mload(o))
+            if iszero(eq(mload(add(u, _PAYMENT_AMOUNT_OFFSET)), 0x112233)) { invalid() }
+            if iszero(eq(mload(add(u, _PAYMENT_MAX_AMOUNT_OFFSET)), 0x223344)) { invalid() }
+            if iszero(eq(mload(add(u, _PAYMENT_PER_GAS_OFFSET)), 0x334455)) { invalid() }
+        }
+    }
```
Vectorized, line 23: A constructor time test. Just in case we update the UserOp in the future and forget to change the offsets.

Vectorized:
Wait, lemme check why the tests are failing.

src/EntryPoint.sol
Vectorized: From a Swiss cheese perspective, I think checking that the caller has a balance of `type(uint256).max` is safer than requiring `msg.sender == address(this)` for the non-reverting workflow. 

It is impossible for any account to have a balance of `type(uint256).max` in practice. 

But a coding mistake might accidentally enable the EntryPoint call this simulate self-call function outside of simulation.

Vectorized:
Ok, ready to merge.

PR #104, opened 2025-04-04
feat: normified non-reverting `simulateExecute`

Vectorized:
This adds 1420 bytes in bytecode size.

Syntax sugar so sinful, it makes me die.

May or may not want to merge, opening this PR to marinate if the syntax sugar is worth it.


Vectorized:
Closing for now.

legion2002 (approved):

PR #113, opened 2025-04-10
feat: make UserOp as PreOp more flexible

Vectorized:
Notes:

- A `nonce` of `type(uint256).max` cannot be used outside of a PreOp.
- An `eoa` of `address(0)` cannot be used outside of a PreOp.
- This is actually safe because the PreOps are included in the EIP712 hash.

If merging this, don't merge #114.


Vectorized:
Closing this. We are using #114.

PR #118, opened 2025-04-11
fix: remove EntryPoint check on initializePREP

Vectorized:


legion2002 (approved):

Vectorized:
Merging. This shouldn't be breaking, as it's a relaxation of requirements.

PR #122, opened 2025-04-15
fix: make simulateExecute directly revert on PaymentError

Vectorized:
So, when `err == PaymentError.selector`, `gCombined` will be left as default 0 -> `paymentOverride = 0` -> final simulation won't revert. 

So the fix is to directly revert if `err == PaymentError.selector`, for the no-revert `simulateExecute` workflow.


legion2002 (approved):

Vectorized:
Let's merge this.

PR #123, opened 2025-04-16
Prototype of a simplified simulate function

legion2002:
This is just experimental atm, to gather feedback about if a simple simulate function like this could work.

**Note:** in production, we would not store the simulation flag in transient storage, instead we'll pass it around in memory to save on TLOAD gas. 

I have removed all the features of the old V1 simulate fun, to demonstrate using the diff, how much scope there is to simplify the codebase here.

### DIFF
--- a/src/EntryPoint.sol
+++ b/src/EntryPoint.sol
@@ -319,13 +319,17 @@ contract EntryPoint is
         }
     }
 
-    function simulateExecuteV2(bytes calldata encodedUserOp) public payable virtual {
+    function simulateExecuteV2(bytes calldata encodedUserOp, uint256 combinedGasOverride)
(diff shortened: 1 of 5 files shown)


src/EntryPoint.sol
```diff
@@ -226,6 +229,9 @@ contract EntryPoint is
     /// This constant is a pun for "chain ID 0".
     uint16 public constant MULTICHAIN_NONCE_PREFIX = 0xc1d0;
 
+    uint256 internal constant SIMULATION_V2_FLAG =
```
Vectorized, line 232: Underscore prefix internals in contracts: `_SIMULATION_V2_FLAG`.

PR #117, opened 2025-04-11
New Paymaster Design with Pre & Post Payment 

legion2002:
### Features
1. Replaces `compensate` with `pay` function, which includes the complete `encodedUserOp` for more flexible paymaster designs.
2. The `pay` function allows the user to break their payment into 2 chunks - pre execution and post execution. If the post execution payment is not completed, the execution batch reverts, but the pre-payment is not returned to the user.
3. General cleanup and refactor. Added "solidity equivalents" and "gas-saved" comment sections over assembly blocks.

#### For ETH payment token: ~33k gas saved over old entrypoint for a Uniswap swap

> ╰─❯ forge snapshot --match-contract Benchmark --diff gas-snapshots/.gas-snapshot-main-eth
> [⠊] Compiling...
> [⠘] Compiling 1 files with Solc 0.8.29
> [⠃] Solc 0.8.29 finished in 7.04s
> Compiler run successful!
> 
> Ran 9 tests for test/Benchmark.t.sol:BenchmarkTest
> [PASS] testERC20TransferDirect() (gas: 29222)
> [PASS] testERC20TransferViaERC4337EntryPoint() (gas: 168709)
> [PASS] testERC20TransferViaERC4337EntryPointWithExtendedCalldata() (gas: 171701)
> [PASS] testERC20TransferViaPortoEntryPoint() (gas: 93354)
> [PASS] testERC20TransferViaPortoEntryPointWithExtendedCalldata() (gas: 95848)
> [PASS] testERC20TransferViaPortoEntryPointWithSpendLimits() (gas: 154372)
> [PASS] testUniswapV2SwapViaERC4337EntryPoint() (gas: 216599)
> [PASS] testUniswapV2SwapViaPortoEntryPoint() (gas: 141257)
> [PASS] test__codesize() (gas: 118137)
> Suite result: ok. 9 passed; 0 failed; 0 skipped; finished in 9.42ms (17.02ms CPU time)
> 
> Ran 1 test suite in 129.20ms (9.42ms CPU time): 9 tests passed, 0 failed, 0 skipped (9 total tests)
> testERC20TransferDirect() (gas: 0 (0.000%))
> testERC20TransferViaERC4337EntryPoint() (gas: 0 (0.000%))
> testERC20TransferViaERC4337EntryPointWithExtendedCalldata() (gas: 0 (0.000%))
> testUniswapV2SwapViaERC4337EntryPoint() (gas: 0 (0.000%))
> test__codesize() (gas: 219 (0.186%))
[...]

PR #128, opened 2025-04-22
Simulate V2 + New compensate design

legion2002:
A version of #117 , that includes the refactored simulate function.

We remove most simulation logic from the entrypoint, and transfer it to a reusable simulator contract.

This provides better security properties, and readability. You can also adjust the step size/algorithms used to get combined gas, by doing offchain analysis of the gas overheads.

We also have the option to add multiple simulation modes using the simulator, without deploying a new entrypoint.

### General Design 

image

We create a new contract called the `Simulator` that can be deployed and maintained completely separately to the Entrypoint.
All simulation logic has been removed from the Entrypoint, and a very basic primitive has been exposed called the `simulateExecute` function, on top of which we can create Simulators with any algorithms we want.
Simulator can have any relay-specific code, without having to update the entrypoint. 

One simulator can work for multiple entrypoints. 

### Relay Changeset
The most relevant change for the relay is - 

>     /// @dev Simulates the execution of a userOp, and finds the combined gas by iteratively increasing it until the simulation passes.
>     /// The start value for combinedGas is gasUsed + original combinedGas.
>     /// Set u.combinedGas to add some starting offset to the gasUsed value.
>     /// @param ep The entry point address
>     /// @param isPrePayment Whether to add gas amount to prePayment or postPayment
>     /// @param paymentPerGas The amount of `paymentToken` to be added per gas unit.
>     /// Total payment is calculated as pre/postPaymentAmount += gasUsed * paymentPerGas.
>     /// @dev Set prePayment or totalPaymentAmount to include any static offset to the gas value.
>     /// @param combinedGasIncrement Basis Points increment to be added for each iteration of searching for combined gas.
>     /// @dev The closer this number is to 10_000, the more precise combined gas will be. But more iterations will be needed.
>     /// @dev This number should always be > 10_000, to get correct results.
>     //// If the increment is too small, the function might run out of gas while finding the combined gas value.
>     /// @param encodedUserOp The encoded user operation
>     /// @return gasUsed The gas used in the successful simulation
>     /// @return combinedGas The first combined gas value that gives a successful simulation.
>     function simulateV1Logs(
>         address ep,
>         bool isPrePayment,
>         uint256 paymentPerGas,
>         uint256 combinedGasIncrement,
>         bytes calldata encodedUserOp
>     ) public payable virtual returns (uint256 gasUsed, uint256 combinedGas)

To maintain 1:1 compatibility with the old simulation function, the relay should call this function with -
isPrepayment = true
paymentPerGas ( same as before ) 
combinedGasIncremeent = 11_000

Note: any offsets for payment or combined gas should be added to the `u.combinedGas` and `u.prePaymentAmount/u.totalPaymentAmount` fields.
It is recommended to initially send u.combinedGas = 10_000 initially, so that small gas overheads caused due to difference in simulation mode/execution mode can be captured.
This static value can be optimized down to 1-2k over time.


src/Simulator.sol
```diff
@@ -0,0 +1,167 @@
...
+
+    function _callEntryPointMemory(
+        address ep,
+        uint256 combinedGasOverride,
+        bytes memory encodedUserOp
+    ) internal {
+        // Set the simulation flag to true
+        assembly ("memory-safe") {
+            let m := mload(0x40)
+            mstore(m, 0x68faa9fd) // function selector of simulateExecute
```
Vectorized, line 56: Assembly comment wise, i'll usually put the full function signature with all the args, so that it is easy to copy-paste the arg to `cast k "<fnSig>"` to check if it's correct.

src/Simulator.sol
```diff
@@ -0,0 +1,167 @@
...
+            u.prePaymentMaxAmount += paymentAmount;
+        }
+
+        u.totalPaymentAmount += paymentAmount;
+        u.totalPaymentMaxAmount += paymentAmount;
+
+        u.combinedGas += gasUsed;
+
+        while (true) {
+            _callEntryPointMemory(ep, 0, abi.encode(u));
```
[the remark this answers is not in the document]
Vectorized: Suggestion:

Make another `_callEntryPointMemory` function that does the `abi.encode` inside. 
But this function caches `mload(0x40)` and restores the at the end of the function.

src/Simulator.sol
```diff
@@ -0,0 +1,167 @@
...
+    function _getGasUsed() internal pure returns (uint256 gasUsed) {
+        assembly ("memory-safe") {
+            let err := shl(224, shr(224, mload(0)))
+            // Check if first 4 bytes are equal to SimulationPassed(uint256)
+            if eq(err, 0x4f0c028c00000000000000000000000000000000000000000000000000000000) {
+                // Execute was successful
+                gasUsed := mload(0x04)
+            }
+        }
+    }
```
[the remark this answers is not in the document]
legion2002: Yeah good catch, I moved the gas used decoding inside the internal function assembly blocks, because I was already returning gasUsed for the isStateOverride flag, for simulateV1Logs.

But there is still a lot of code dupliication between the 2 functions, any ideas on how to improve this?

src/Simulator.sol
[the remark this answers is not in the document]
legion2002: Yup, good point. Something like this work? - https://github.com/ithacaxyz/account/pull/128/commits/2ab512070d4cb3c00bf2450dc4f5b1e9f5fe33f7

src/interfaces/ICommon.sol
```diff
@@ -0,0 +1,103 @@
...
+        uint256 totalPaymentMaxAmount;
+        /// @dev The combined gas limit for payment, verification, and calling the EOA.
+        uint256 combinedGas;
+        /// @dev Optional data for `initPREP` on the delegation.
+        /// This is encoded using ERC7821 style batch execution encoding.
+        /// (ERC7821 is a variant of ERC7579).
+        /// `abi.encode(calls, abi.encodePacked(bytes32(saltAndDelegation)))`,
+        /// where `calls` is of type `Call[]`,
+        /// and `saltAndDelegation` is `bytes32((uint256(salt) << 160) | uint160(delegation))`.
+        bytes initData;
```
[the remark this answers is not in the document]
legion2002: Yup you're right. Moved.

src/EntryPoint.sol
```diff
@@ -318,178 +216,42 @@ contract EntryPoint is EIP712, Ownable, CallContextChecker, ReentrancyGuardTrans
...
+        bytes calldata encodedUserOp
+    ) external payable returns (uint256) {
+        // If Simulation Fails, then it will revert here.
+        (uint256 gUsed, bytes4 err) = _execute(encodedUserOp, combinedGasOverride, 1);
+
+        if (err != 0) {
             assembly ("memory-safe") {
-                revert(add(0x20, result), mload(result))
+                mstore(0x00, shl(224, err))
+                revert(0x00, 0x20)
```
Vectorized, line 242: `bytes4 err` is left-aligned under the hood `0x1122334400000000000000000000000000000000000000000000000000000000`.

`uint32 err` is right-aligned under the hood `0x0000000000000000000000000000000000000000000000000000000011223344`.

So the correct code is:
```
mstore(0x00, err)
revert(0x00, 0x20)
```

src/EntryPoint.sol
```diff
@@ -524,57 +286,83 @@ contract EntryPoint is EIP712, Ownable, CallContextChecker, ReentrancyGuardTrans
...
+        // unchecked {
+        //     // Check if there's sufficient gas left for the gas-limited self calls
+        //     // via the 63/64 rule. This is for gas estimation. If the total amount of gas
+        //     // for the whole transaction is insufficient, revert.
+        //     if (((gasleft() * 63) >> 6) < Math.saturatingAdd(g, _INNER_GAS_OVERHEAD)) {
+        //         if(simulationFlags != 1) {
+        //             revert InsufficientGas();
+        //         }
+        //     }
+        // }
```
Vectorized, line 331: The 63/64 rule was included in the original EntryPoint for anti-griefing.

Without it:
- an attacker can intentionally make the call to this EntryPoint at a very deep depth, and make the `execute` on the Delegation fail.
- an attacker can provide just enough gas to make the `execute` on the Delegation run out-of-gas, but leave enough gas for the remaining operations after the `execute` to complete.

Vectorized: i think even if users trust the relayers, and even if the relayers are in a private mempool, this might still be needed.

Tbh, I'll just add it so that we don't need to fight the FUD of not being censorship resistant.

src/EntryPoint.sol
legion2002: good catch! fixed this now. 
It wasn't caught in the tests, because it is technically not possible to reach this code path, since we explicitly revert in simulation mode.

I just keep it here, as a failsafe against us forgetting to revert for a non zero error selector.

src/EntryPoint.sol
gakonst: >An attacker can intentionally make the call to this EntryPoint at a very deep depth, and make the execute on the Delegation fail (cuz the EVM has a max call depth of 1024). The idea is that with a _INNER_GAS_OVERHEAD, if the call is deep enough (about 50 depth-ish), the remaining gas will be sufficiently decayed to not pass this check.

Can we add a unit test pls for this?

PR #132, opened 2025-04-22
Optimize _isSuperAdmin

Vectorized:
Shaves a bit of gas. Better than nothing.

This optimization will be more significant if the key is quite long (e.g. WebAuthn keys).

Not breaking.


legion2002 (approved):

Vectorized:
Merging this, it is non-breaking. Doesn't change the layout.

PR #141, opened 2025-04-30
mcopy instead abi.encode

fico23:


src/GuardedExecutor.sol
```diff
@@ -569,7 +569,11 @@ abstract contract GuardedExecutor is ERC7821 {
 
     /// @dev Stores the spend struct.
     function _storeSpend(LibBytes.BytesStorage storage $, TokenPeriodSpend memory spend) internal {
-        LibBytes.set($, LibZip.cdCompress(abi.encode(spend)));
+        bytes memory encoded = new bytes(96);
+        assembly ("memory-safe") {
+            mcopy(add(encoded, 0x20), spend, 96)
+        }
+        LibBytes.set($, LibZip.cdCompress(encoded));
```
Vectorized, line 576:
```suggestion
        bytes memory encoded;
        assembly ("memory-safe") {
            // This is only safe if `spend` comes from `_loadSpend`, which fits our usage.
            encoded := sub(spend, 0x20)
        }
        LibBytes.set($, LibZip.cdCompress(encoded));
```

Vectorized (commented):
I think the gas saved is not enough for the extra SLOCs. 

Possibly optimal, but as you can see it's not pretty:

```solidity
bytes memory encoded;
uint256 cached;
assembly ("memory-safe") {
    encoded := sub(spend, 0x20)
    cached := mload(encoded)
    mstore(encoded, 0x60)
}
LibBytes.set($, LibZip.cdCompress(encoded));
assembly ("memory-safe") {
    mstore(encoded, cached)
}
```

can save nearly 1k ish gas, but not sure if the extra SLOCs are worth it.

Vectorized:
Might close this until I feel the urge to optimize.

Vectorized:
Closing this first. Might re-open if we need that small gas shave.

Issue #140, opened 2025-04-29
Stateless Entrypoint + Standardized nonce flags


legion2002:
### Storage of Entrypoint

>     /// @dev Holds the storage.
>     struct EntryPointStorage {
>         /// @dev Mapping of (`eoa`, `seqKey`) to nonce sequence.
>         /// We use a `LibStorage.Ref` instead of a uint64 for performance.
>         mapping(address => mapping(uint192 => LibStorage.Ref)) nonceSeqs;
>         /// @dev Mapping of (`eoa`, `nonce`) to the error selector.
>         /// If `uint64(nonce) < nonceSeqs[eoa][uint192(nonce >> 64)]`,
>         /// it means that the nonce has either been used or invalidated,
>         /// and a non-zero error selector denotes an error.
>         /// Otherwise, if `uint64(nonce) >= nonceSeqs[eoa][uint192(nonce >> 64)]`,
>         /// we would expect that the error selector is zero (i.e. uninitialized).
>         mapping(address => mapping(uint256 => bytes4)) errs;
>         /// @dev A bitmap to mark ERC7683 order IDs as filled, to prevent filling replays.
>         LibBitmap.Bitmap filledOrderIds;
>     }

* `errs` mapping is legacy, not used anywhere in the codebase rn, so this is going to be removed.
* `filledOrderIds` is only used for the ERC 7683 fill function, which we also never use.
* `nonceSeqs` used to maintain nonces in the EP for userOps. Delegation does not check for nonces if call is coming from EP. ( More on this below )

### Current Nonce Design 

There are 2 main ways in which we use nonce based flags - 
1. If nonce starts with `0xc1d0`, then we consider the multichain prefix in the entrypoint, and don't include the chainId in the 712.
2. If `nonce == type(uint256).max`, then we skip the nonce check altogether for preOps
Flag 2, only works for preOps, it doesn't work for userOps or direct execute calls to the delegation.

And for the entrypoint and direct execute calls, there is no way to skip the nonce check completely. This prevents some use cases where we want the execution to be replayable.

### Problems
1. This set of rules is confusing to implement for integrators, as different execution types treat the nonce separately. There should be a standard expected behaviour for all nonces.
2. We need to have storage in entrypoint just to maintain nonces in the EP for userOps and preOps, when we could be doing these nonce checks all in one place in the Delegation.
[...]

PR #142, opened 2025-04-30
Make Entrypoint completely stateless

legion2002:
Implements some of the features mentioned in #140.

* Entrypoint is now completely stateless, userOp nonce validation logic has been moved to the Delegation.
* Other unused functions like fill etc. have also been removed. 
* Costs 1k extra gas, because of an additional call to the `checkAndIncrementNonce` function in the delegation by the EP. We can remove this gas overhead by grouping this function along with other calls like `execute` and `pay`, but it adds complications for paymaster flows, and increases code complexity.
* All call types now share a nonce. So a userOp with nonce `0` will invalidate a direct delegation call bundle with nonce `0`. I think this is desirable, to avoid user mistakes while signing things. Separate `seqKeys` should be used, for separate execution types.
* `getNonce` should be called from the delegation instead of the entrypoint

Note: Since the discussion about optional `nonce` and `from` field has not been resolved yet, I have removed the optional nonces from PreOps and UserOps. If we choose to add this at some point, it is now easier to do so in a standardized way inside the Delegation.


src/EntryPoint.sol
```diff
@@ -660,64 +605,25 @@ contract EntryPoint is
...
-        payable
-        virtual
-        returns (bytes4)
-    {
-        if (orderId != bytes32(0)) {
-            if (!_getEntryPointStorage().filledOrderIds.toggle(uint256(orderId))) {
-                revert OrderAlreadyFilled();
+            if iszero(call(gas(), eoa, 0, add(m, 0x1c), 0x24, 0x00, 0x00)) {
+                mstore(0x00, shl(224, 0x756688fe)) // `InvalidNonce()`.
+                revert(0x00, 0x20)
```
Vectorized, line 622: For this call, i’d use the scratch space 

```solidity
mstore(0x00, 0x9e49fbf1)
mstore(0x20, nonce)
if iszero(call(gas(), eoa, 0, 0x1c, 0x24, 0x00, 0x00)) …
```

src/Delegation.sol
```diff
@@ -307,6 +307,14 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
         emit NonceInvalidated(nonce);
     }
 
+    /// @dev Checks current nonce and increments the sequence for the `seqKey`.
+    function checkAndIncrementNonce(uint256 nonce) public virtual {
```
Vectorized, line 311: Make this payable.

src/EntryPoint.sol
Vectorized: Also, for reverting, i’ll always use
```solidity
mstore(0x00, 0x736688fe) // Ain’t no need to shl 224. Save bytecode size. Save gas.
revert(0x1c, 0x04) // This is exactly 4 bytes.
```

legion2002: Cool suggestions, implemented all!

Vectorized (approved):

Vectorized:
Not sure why remove the ability for the EntryPoint to accure tokens tho (for us). It’s a precious warm address.

PR #143, opened 2025-04-30
chore: update payment in combinedGas loop

legion2002:
We do this to cover edge cases in paymaster flows, where a small change in the payment amount can cause big codepath changes, causing simulation to fail.

### DIFF
--- a/src/Simulator.sol
+++ b/src/Simulator.sol
@@ -25,10 +25,10 @@ contract Simulator {
     function _updatePaymentAmounts(
         ICommon.UserOp memory u,
         bool isPrePayment,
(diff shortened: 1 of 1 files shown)


src/Simulator.sol
```diff
@@ -176,6 +176,8 @@ contract Simulator {
                 return (gasUsed, u.combinedGas);
             }
 
+            _updatePaymentAmounts(u, isPrePayment, gasUsed, paymentPerGas);
```
klkvr, line 179: shouldn't this use the delta between old and new gas estimate instead?

klkvr (approved):

PR #146, opened 2025-05-05
feat: upgradeHook

Vectorized:
Marinating if this is generalized enough and also secure.

This is non-breaking, as of now.

### DIFF
--- a/src/Delegation.sol
+++ b/src/Delegation.sol
@@ -338,7 +338,7 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
         // Using a dedicated guard makes the hook only callable via this function
         // prevents direct self-calls which may accidentally use the wrong hook ID and version.
         LibTransient.tBytes32(_UPGRADE_HOOK_GUARD_TRANSIENT_SLOT).set(_UPGRADE_HOOK_ID);
-        // We MUST use `this`, so that it uses the new implementation's `updgradeHook`.
+        // We MUST use `this`, so that it uses the new implementation's `upgradeHook`.
         require(this.upgradeHook(LibString.toSmallString(version)));
     }
 
--- a/src/Delegation.sol
+++ b/src/Delegation.sol
@@ -345,6 +345,8 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
     /// @dev For this very first version, the upgrade hook is just an no-op.
     /// For future implementations, we will have an upgrade hook which can contain logic
     /// to migrate storage on a case-by-case basis if needed.
+    /// If this hook is implemented to mutate storage,
+    /// it MUST check that `_UPGRADE_HOOK_GUARD_TRANSIENT_SLOT` is correctly set.
     function upgradeHook(bytes32 previousVersion) public virtual returns (bool) {
         previousVersion = previousVersion; // Silence unused variable warning.
         return true; // Always returns true for cheaper success check (even with plain Solidity).
--- a/src/Delegation.sol
+++ b/src/Delegation.sol
@@ -201,6 +201,14 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
     /// This constant is a pun for "chain ID 0".
     uint16 public constant MULTICHAIN_NONCE_PREFIX = 0xc1d0;
 
+    /// @dev A unique identifier to be passed into `upgradeHook(bytes32,string)`.
+    bytes32 internal constant _UPGRADE_HOOK_ID = keccak256("PORTO_DELEGATION_UPGRADE_HOOK_ID");
+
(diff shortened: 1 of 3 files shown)


test/utils/mocks/MockSampleDelegateCallTarget.sol
```diff
@@ -21,4 +28,14 @@ contract MockSampleDelegateCallTarget {
     function revertWithData(bytes memory data) public pure {
         revert ErrorWithData(data);
     }
+
+    function upgradeHook(bytes32 hookId, string calldata) public returns (bool) {
+        require(hookId == _UPGRADE_HOOK_ID);
```
gakonst, line 33: is the `tload` guard basically a cheaper `require msg.sender == address(this)`?

src/Delegation.sol
```diff
@@ -323,6 +331,21 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
     /// only a new EIP-7702 transaction can change the authority's logic.
     function upgradeProxyDelegation(address newImplementation) public virtual onlyThis {
         LibEIP7702.upgradeProxyDelegation(newImplementation);
+        (, string memory version) = _domainNameAndVersion();
+        bytes memory data =
+            abi.encodeWithSignature("upgradeHook(bytes32,string)", _UPGRADE_HOOK_ID, version);
+        assembly ("memory-safe") {
+            tstore(_UPGRADE_HOOK_GUARD_TRANSIENT_SLOT, 1)
+            mstore(0x00, 0) // Zeroize the return slot.
+            let success := delegatecall(gas(), newImplementation, add(0x20, data), data, 0x00, 0x20)
```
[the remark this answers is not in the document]
Vectorized: Ah… good call (pun intended).

I think this patten, which was inspired by OZ’s UUPSUpgradable doesn’t matter here. Let’s make it `call` for more safety.

test/utils/mocks/MockSampleDelegateCallTarget.sol
Vectorized: Once we’ve made it into a `call`, we can use `msg.sender == address(this)`. 
Cuz `delegatecall` doesn’t change `msg.sender`.

Lemme think if there’s any merit in blocking directly accessing from `execute`.

Vectorized: Ok, I think let's keep a transient storage guard for swiss cheese model.

src/Delegation.sol
```diff
@@ -201,6 +204,14 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
     /// This constant is a pun for "chain ID 0".
     uint16 public constant MULTICHAIN_NONCE_PREFIX = 0xc1d0;
 
+    /// @dev A unique identifier to be passed into `upgradeHook(bytes32 previousVersion)`
+    /// via the transient storage slot at `_UPGRADE_HOOK_GUARD_TRANSIENT_SLOT`.
+    bytes32 internal constant _UPGRADE_HOOK_ID = keccak256("PORTO_DELEGATION_UPGRADE_HOOK_ID");
+
+    /// @dev This transient slot must be set to `_UPGRADE_HOOK_ID` before `upgradeHook` can be processed.
+    bytes32 internal constant _UPGRADE_HOOK_GUARD_TRANSIENT_SLOT =
+        bytes32(uint256(keccak256("_UPGRADE_HOOK_GUARD_TRANSIENT_SLOT")) - 1);
```
Vectorized, line 213: Trivia: this may not be actually evaluated on compile time (yes, it's so dumb).

But since the upgrade function is not often called, i think we can deal with some more readable Solidity instead of a hardcoded constant.

src/Delegation.sol
```diff
@@ -323,6 +334,23 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
...
+        require(this.upgradeHook(LibString.toSmallString(version)));
+    }
+
+    /// @dev For this very first version, the upgrade hook is just an no-op.
+    /// Provided to enable calling it via plain Solidity.
+    /// For future implementations, we will have an upgrade hook which can contain logic
+    /// to migrate storage on a case-by-case basis if needed.
+    /// If this hook is implemented to mutate storage,
+    /// it MUST check that `_UPGRADE_HOOK_GUARD_TRANSIENT_SLOT` is correctly set.
+    function upgradeHook(bytes32 previousVersion) public virtual returns (bool) {
```
legion2002, line 351: We can make this external, to enforce that this is always called with `this.upgradeHook`.

And even though this is a no-op, should we add an `onlyThis` and transient slot check as a template to make sure we don't forget this for future upgrades.

legion2002 (approved):

PR #147, opened 2025-05-05
feat: Allow Porto Delegation accounts to act as Paymasters

legion2002:
Allows delegation to be the paymaster.

@relay Note that after this PR, if you explicitly set the payer field to the EOA itself for normal flows, then it will give an error, because it will consider the EOA to be a paymaster.

For non-paymaster flows, u.payer should just be left empty.

### DIFF
--- a/src/Delegation.sol
+++ b/src/Delegation.sol
@@ -548,7 +548,7 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
 
             if (!isValid) {
                 // The error is skipped, if the paymaster is in simulation mode.
-                /// @dev to simulate a paymaster, state override the address of the delegation
+                /// @dev to simulate a paymaster, state override the address of the entrypoint
                 /// to type(uint256).max
                 revert Unauthorized();
             }
--- a/src/Delegation.sol
+++ b/src/Delegation.sol
@@ -70,6 +70,8 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
 
     /// @dev Holds the storage.
     struct DelegationStorage {
+        /// @dev 32 byte word for bit flags, can be extended to add features in future upgrades.
+        uint256 flags;
         /// @dev The label.
         LibBytes.BytesStorage label;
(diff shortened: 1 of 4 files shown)


jxom (approved):

src/EntryPoint.sol
```diff
@@ -642,7 +642,7 @@ contract EntryPoint is IEntryPoint, EIP712, CallContextChecker, ReentrancyGuardT
 
         // Call the pay function on the delegation contract
         // Equivalent Solidity code:
-        // IDelegation(payer).pay(paymentAmount, keyHash, abi.encode(u));
+        // IDelegation(payer).pay(paymentAmount, keyHash, digest, abi.encode(u));
```
Vectorized, line 645: Nice comment catch. Ty

src/Delegation.sol
```diff
@@ -528,10 +528,32 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
...
+                isValid = true;
+            }
+
+            if (!isValid) {
+                // The error is skipped, if the paymaster is in simulation mode.
+                /// @dev to simulate a paymaster, state override the address of the entrypoint
+                /// to type(uint256).max
+                revert Unauthorized();
+            }
+        }
```
Vectorized, line 555: Comment nit (cuz it’s not strictly a flag imo):

```suggestion
        // If this delegation is the paymaster, validate the paymaster signature.
        if (userOp.payer == address(this)) {
            (bool isValid,) = unwrapAndValidateSignature(userOpDigest, userOp.paymentSignature);

            // If this is a simulation, signature validation errors are skipped.
            if (address(ENTRY_POINT).balance == type(uint256).max) {
                isValid = true;
            }

            if (!isValid) {
                // The error is skipped, if the paymaster is in simulation mode.
                /// @dev to simulate a paymaster, state override the address of the entrypoint
                /// to type(uint256).max
                revert Unauthorized();
            }
        }
```

Vectorized (approved):

src/Delegation.sol
legion2002: Good catch, resolved!

Issue #137, opened 2025-04-26
Make accounts pausable, with an optional pausing authority


legion2002:
- People can easily opt out of pausability, by setting the pause authority to 0.
- Down the line people can pay for AI monitoring services which pause their accounts if they see a widespread attack, or a suspicious transaction related to the user's account in the mempool.
- Tight pack the pause bit with some hot storage slot, so that it doesn't add to gas costs for general user flows.

PR #149, opened 2025-05-07
Add global pause in the Entrypoint

legion2002:
PR for  #137 

@relay We SHOULD update the `pauseAuthority` after the initial contracts deployment to a safe key, by calling `setPauseAuthority`. 

In the event of an attack. The `pauseAuthority` needs to call the `pause` function in the entrypoint, which will pause all delegation accounts for upto a month.

In the current design, users cannot opt out of this.

**Benchmarks: ** Adds 2.5k extra gas for all normal flows.


src/Delegation.sol
```diff
@@ -94,6 +94,12 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
         EnumerableSetLib.AddressSet approvedImplementations;
         /// @dev Mapping of approved implementations to their callers storage.
         mapping(address => LibStorage.Bump) approvedImplementationCallers;
+        /// @dev flags bitmap, which can be extended in the future to add boolean values.
+        LibBitmap.Bitmap flags;
+        /// @dev Address that can call the `pause` function.
+        address pauseAuthority;
+        /// @dev Last pause timestamp.
+        uint256 lastPauseTimestamp;
```
Vectorized, line 102: Since these two fields will always be together, make `lastPauseTimestamp` into a uint40.

Vectorized: I think calling it `lastPaused` feels more concise.

src/Delegation.sol
```diff
@@ -178,6 +187,12 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
     /// on the Delegation with a `keyHash`, bypassing the EntryPoint.
     event NonceInvalidated(uint256 nonce);
 
+    /// @dev The pause flag has been updated.
+    event PauseUpdated(bool indexed isPaused);
```
Vectorized, line 191: `PauseSet`, shorter.

src/Delegation.sol
```diff
@@ -204,6 +219,9 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
     /// This constant is a pun for "chain ID 0".
     uint16 public constant MULTICHAIN_NONCE_PREFIX = 0xc1d0;
 
+    /// @dev Time period after which the contract can be unpaused by anyone.
+    uint256 public constant PAUSE_EXPIRY = 4 weeks;
```
Vectorized, line 223: `PAUSE_TIMEOUT`? 

Somehow `PAUSE_EXPIRY` feels like it's the exact timestamp when people can unpause.

src/Delegation.sol
```diff
@@ -216,6 +234,9 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
     /// to prevent off-chain full enumeration from running out-of-gas.
     uint256 internal constant _CAP = 512;
 
+    /// @dev Bit position of the pause flag.
+    uint256 internal constant PAUSE_FLAG = 0;
```
Vectorized, line 238: `_PAUSE_FLAG`.

src/Delegation.sol
```diff
@@ -358,6 +384,44 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
         return true;
     }
 
+    /// @dev Can be used to pause/unpause the contract, in case of emergencies.
+    /// - Pausing the contract will make all signature validations fail,
+    ///   effectively blocking all pay, execute, isValidSignature attempts.
```
Vectorized, line 389: To be precise, the pausing functionality only affects workflows where `unwrapAndValidateSignature` is used.

The `execute` workflow where the EOA executes itself won't be blocked. This is **ok**, since this workflow is very simple, well-understood, and is mostly based on the audited Solady's 7821.

src/Delegation.sol
```diff
@@ -358,6 +384,44 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
...
+        DelegationStorage storage $ = _getDelegationStorage();
+        if (isPause) {
+            // Owners can use this 1 week buffer, to update their `pauseAuthority` if needed.
+            if (
+                msg.sender != $.pauseAuthority
+                    || block.timestamp < $.lastPauseTimestamp + PAUSE_EXPIRY + 1 weeks
+                    || $.flags.get(PAUSE_FLAG)
+            ) {
+                revert Unauthorized();
+            }
```
Vectorized, line 405: Imo, `$.lastPauseTimestamp + PAUSE_EXPIRY` into a temp variable `pauseUntil`. Then `forge fmt` will hopefully make this into a single line.

src/PauseAuthority.sol
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    function setPauseAuthority(address newPauseAuthority) public virtual {
+        (address authority, uint40 lastPaused) = getPauseConfig();
+        if (msg.sender != authority) {
+            revert Unauthorized();
+        }
+
+        // TODO: Check if solidity sanitizes the newPauseAuthority address already
+        _pauseConfig = lastPaused << 160 | (uint160(newPauseAuthority) & type(uint160).max);
```
Vectorized, line 71: imo, the neatest way is:

`(uint256(lastPaused) << 160) | uint160(newPauseAuthority)`

When you upcast a `uint160` to `uint256`, the upper bits are auto-cleared. The `uint256 | uint160` will upcast the right.

Also, note that `uint40(x) << 96` gives 0 in chisel, so do explicity upcast the `lastPaused`.

src/PauseAuthority.sol
```diff
@@ -0,0 +1,74 @@
...
+                revert Unauthorized();
+            }
+        }
+
+        emit PauseSet(isPause);
+    }
+
+    /// @dev Returns the pause authority and the last pause timestamp.
+    function getPauseConfig() public view virtual returns (address, uint40) {
+        return (address(uint160(_pauseConfig & type(uint160).max)), uint40(_pauseConfig >> 160));
```
Vectorized, line 61:
```suggestion
        return (address(uint160(_pauseConfig)), uint40(_pauseConfig >> 160));
```

Solidity will lazily auto-clean upper bits if necessary, and will do so when the variables are returned to the external call.

src/EntryPoint.sol
```diff
@@ -133,6 +141,14 @@ contract EntryPoint is IEntryPoint, EIP712, CallContextChecker, ReentrancyGuardT
     /// Should be enough for a cold zero to non-zero SSTORE + a warm SSTORE + a few SLOADs.
     uint256 internal constant _REFUND_GAS = 50000;
 
+    ////////////////////////////////////////////////////////////////////////
+    // Constructor
+    ////////////////////////////////////////////////////////////////////////
+
+    constructor(address pauseAuthority) {
+        _pauseConfig = uint160(pauseAuthority) & type(uint160).max;
```
Vectorized, line 149:
```suggestion
        _pauseConfig = uint160(pauseAuthority);
```

When a uint160 is upcasted to an uint256, the upper bits are cleaned.

Vectorized (approved):

Issue #151, opened 2025-05-07
Add multisig keytype in Delegation


legion2002:
Multi-sig keys are important for many important use cases. 

We can add this key type with some kind of recursion over other key types. 

Reference - https://github.com/gakonst/simple-multisig/blob/master/contracts/SimpleMultiSig.sol#L13-L94

PR #153, opened 2025-05-12
Add External key type + multi sig

legion2002:
### Features
- Adds support for `External` keytype. Which allows external contracts that implement the `ISigner` 
  to verify signatures. 
- Adds a `MultiSigSigner` which adds multi-sig support to the account, using `External` keytype.#151 
- Adds a stack of `context key hashes` to the account. External observers can check which key signed
  the account's current execution, by calling the `getContextKeyHash` function on the msg.sender.
  This has a variety of usecases like validating key operations, setting data in account registry etc.
  This mechanism can also be used to solve #154 
- Only adds features. Should not have breaking changes for @relay.

### DIFF
--- a/src/Delegation.sol
+++ b/src/Delegation.sol
@@ -402,6 +402,8 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
         }
     }
 
+    /// @dev Return the key hash that signed the latest execution context.
+    /// @dev Returns bytes32(0) if the EOA key was used.
     function getContextKeyHash() public view virtual returns (bytes32) {
         LibTStack.TStack memory t = LibTStack.tStack(_KEYHASH_STACK_TRANSIENT_SLOT);
         if (LibTStack.size(t) == 0) {
--- a/src/Delegation.sol
+++ b/src/Delegation.sol
@@ -403,7 +403,12 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
     }
 
     function getContextKeyHash() public view virtual returns (bytes32) {
-        return LibTStack.TStack(_KEYHASH_STACK_TRANSIENT_SLOT).top();
+        LibTStack.TStack memory t = LibTStack.tStack(_KEYHASH_STACK_TRANSIENT_SLOT);
+        if (LibTStack.size(t) == 0) {
+            return bytes32(0);
+        }
+
+        return LibTStack.top(t);
     }
 
     /// @dev Returns the hash of the key, which does not includes the expiry.
--- a/src/Delegation.sol
+++ b/src/Delegation.sol
@@ -44,7 +44,7 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
         P256,
         WebAuthnP256,
         Secp256k1,
-        MultiSig
+        External
     }
 
     /// @dev A key that can be used to authorize call.
@@ -70,8 +70,6 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
         /// @dev The `msg.senders` that can use `isValidSignature`
         /// to successfully validate a signature for a given key hash.
         EnumerableSetLib.AddressSet checkers;
-        /// @dev Arbitrary data for any key type.
-        bytes data;
     }
 
     /// @dev Holds the storage.
@@ -131,6 +129,9 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
 
     /// @dev The `keyType` cannot be super admin.
     error KeyTypeCannotBeSuperAdmin();
+
+    /// @dev The public key is invalid.
+    error InvalidPublicKey();
 
     ////////////////////////////////////////////////////////////////////////
     // Events
@@ -590,50 +591,13 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
                 digest = EfficientHashLib.sha2(digest); // `sha256(abi.encode(digest))`.
             }
         }
+
         Key memory key = getKey(keyHash);
 
         // Early return if the key has expired.
         if (LibBit.and(key.expiry != 0, block.timestamp > key.expiry)) return (false, keyHash);
 
-        if (key.keyType == KeyType.MultiSig) {
-            bytes memory keyData = _getKeyExtraStorage(keyHash).data;
-            bytes[] memory signatures = abi.decode(signature, (bytes[]));
-            (uint256 threshold, bytes32[] memory multiSigKeys) =
-                abi.decode(keyData, (uint256, bytes32[]));
-
-            uint256 validKeyNum;
-
-            for (uint256 i; i < signatures.length; ++i) {
-                // temporary self call to make this work with a memory array
-                (bool v, bytes32 k) = this.unwrapAndValidateSignature(digest, signatures[i]);
-
-                if (!v) {
-                    return (false, keyHash);
-                }
-
-                uint256 j;
-                while (j < multiSigKeys.length) {
-                    if (multiSigKeys[j] == k) {
-                        validKeyNum++;
-                        multiSigKeys[j] = bytes32(0);
-                    }
-
-                    if (validKeyNum == threshold) {
-                        return (true, keyHash);
-                    }
-
-                    j++;
-                }
-
-                // This means that the keyHash was not found
-                if (j == multiSigKeys.length) {
-                    return (false, keyHash);
-                }
-            }
-
-            // If we reach here, then the required threshold was not met.
-            return (false, keyHash);
-        } else if (key.keyType == KeyType.P256) {
+        if (key.keyType == KeyType.P256) {
             // The try decode functions returns `(0,0)` if the bytes is too short,
             // which will make the signature check fail.
             (bytes32 r, bytes32 s) = P256.tryDecodePointCalldata(signature);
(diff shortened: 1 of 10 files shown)


src/MultiSigSigner.sol
```diff
@@ -0,0 +1,146 @@
...
+        Config memory config = configs[msg.sender][keyHash];
+
+        uint256 validKeyNum;
+
+        for (uint256 i; i < signatures.length; ++i) {
+            (bool v, bytes32 k) =
+                IDelegation(msg.sender).unwrapAndValidateSignature(digest, signatures[i]);
+
+            if (!v) {
+                return FAIL_VALUE;
```
gakonst, line 115: no 1 letter variable names

src/Delegation.sol
```diff
@@ -652,6 +658,29 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
             isValid = SignatureCheckerLib.isValidSignatureNowCalldata(
                 abi.decode(key.publicKey, (address)), digest, signature
             );
+        } else if (key.keyType == KeyType.External) {
+            // The public key of an external key type HAS to be 32 bytes.
+            // Top 20 bytes: address of the signer.
+            // Bottom 12 bytes: arbitrary data, that should be used as a salt.
```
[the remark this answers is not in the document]
legion2002: The keyhash for any key in an account is calculated as keccak(keytype,publickey)
If we set the public key, to only be the address of a signer, without a salt. That means we can only add 1 key for 1 signer to an account. 

For eg.
There is a joint account between 2 institutions. Each institution has a multi sig key with different executives.
MultiSigSigner is deployed at 0xabc address

If we don't have this salt, then the first institution adds a key. The public key is 0xabc. Keyhash is keccak(Keytype.External,0xabc)

The next institution tries to add their multi sig to the same account.
They also have the public key 0xabc, so they also get the same keyHash. So instead of creating a new key, they override the old multisig

src/MultiSigSigner.sol
```diff
@@ -0,0 +1,179 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.23;
+
+import {IDelegation} from "./interfaces/IDelegation.sol";
+import {ISigner} from "./interfaces/ISigner.sol";
+
```
Vectorized, line 6: Note to self:

This is a singleton signer that enables many EOAs to maintain many multisig keys.

src/MultiSigSigner.sol
```diff
@@ -0,0 +1,179 @@
...
+    ////////////////////////////////////////////////////////////////////////
+    // Constants
+    ////////////////////////////////////////////////////////////////////////
+
+    /// @dev The magic value returned by `isValidSignatureWithKeyHash` when the signature is valid.
+    /// - Calcualated as: bytes4(keccak256("isValidSignatureWithKeyHash(bytes32,bytes32,bytes)")
+    bytes4 internal constant MAGIC_VALUE = 0x8afc93b4;
+
+    /// @dev The magic value returned by `isValidSignatureWithKeyHash` when the signature is invalid.
+    bytes4 internal constant FAIL_VALUE = 0xffffffff;
```
Vectorized, line 17: Nit:

Underscore prefix internal constants on contracts.

src/MultiSigSigner.sol
```diff
@@ -0,0 +1,179 @@
...
+    error ConfigAlreadySet();
+
+    ////////////////////////////////////////////////////////////////////////
+    // Storage
+    ////////////////////////////////////////////////////////////////////////
+
+    struct Config {
+        uint256 threshold;
+        bytes32[] ownerKeyHashes;
+    }
```
Vectorized, line 42: Can be packed, but since this multisig isn't often used, it's ok.

src/MultiSigSigner.sol
```diff
@@ -0,0 +1,179 @@
...
+
+    ////////////////////////////////////////////////////////////////////////
+    // Config Functions
+    ////////////////////////////////////////////////////////////////////////
+
+    function _checkKeyHash(bytes32 expectedKeyHash) internal view {
+        bytes32 keyHash = IDelegation(msg.sender).getContextKeyHash();
+        if (keyHash != expectedKeyHash) revert InvalidKeyHash();
+    }
```
Vectorized, line 64: Note to self:

This expects that the caller is an account that implements the `getContextKeyhash()`.

Safe from spoofing.

src/MultiSigSigner.sol
```diff
@@ -0,0 +1,179 @@
...
+
+        Config storage config = _configs[msg.sender][keyHash];
+
+        if (config.threshold > 0) {
+            revert ConfigAlreadySet();
+        }
+
+        _configs[msg.sender][keyHash] =
+            Config({threshold: threshold, ownerKeyHashes: ownerKeyHashes});
+    }
```
Vectorized, line 80: Prefer renaming to `initConfig`, to highlight that this is not an update function.

src/MultiSigSigner.sol
```diff
@@ -0,0 +1,179 @@
...
+
+    function setThreshold(bytes32 keyHash, uint256 threshold) public {
+        _checkKeyHash(keyHash);
+
+        Config storage config = _configs[msg.sender][keyHash];
+
+        if (threshold == 0 || threshold > config.ownerKeyHashes.length) revert InvalidThreshold();
+
+        config.threshold = threshold;
+    }
```
Vectorized, line 121: Note to self:

The `keyhash` on `msg.sender` which created the config is authorized to change the threshold.

src/MultiSigSigner.sol
```diff
@@ -0,0 +1,179 @@
...
+    /// - This will iteratively make a call to the address(msg.sender).unwrapAndValidateSignature
+    ///   for each owner key hash in the config.
+    /// - Signature of a multi-sig should be encoded as abi.encode(bytes[] memory ownerSignatures)
+    /// - For efficiency, place the signatures in the same order as the ownerKeyHashes in the config.
+    /// - Failing owner signatures are ignored, as long as valid signaturs > threshold.
+    function isValidSignatureWithKeyHash(bytes32 digest, bytes32 keyHash, bytes memory signature)
+        public
+        view
+        returns (bytes4 magicValue)
+    {
```
Vectorized, line 137: Prefer if `keyHash` is abi.encoded into `signature` too, then we can use ERC-1271 style.

I think there's some assembly trick to do that without decoding and re-encoding on the Delegation..

src/MultiSigSigner.sol
```diff
@@ -0,0 +1,179 @@
...
+        bytes[] memory signatures = abi.decode(signature, (bytes[]));
+        Config memory config = _configs[msg.sender][keyHash];
+
+        uint256 validKeyNum;
+
+        for (uint256 i; i < signatures.length; ++i) {
+            (bool isValid, bytes32 ownerKeyHash) =
+                IDelegation(msg.sender).unwrapAndValidateSignature(digest, signatures[i]);
```
Vectorized, line 145: Note to self:

This MultiSigSigner requires the `msg.sender` to implement Delegation.

src/MultiSigSigner.sol
Vectorized: Let's leave it as it is. 

`isValidSignatureWithKeyHash` is intentionally made to not be ERC-1271, to highlight that this contract depends on `msg.sender` being the porto Delegation.

Vectorized (commented):
Need to add more comments, especially Natspec.

Vectorized:
@legion2002 #167

src/MultiSigSigner.sol
legion2002: Yeah it was a deliberate decision to make this non standard. As we don't want people to call the Signers directly.
They should always go through the delegation. Because delegation has important checks like `isSuperAdmin` etc.

If the Signer really wants to be 1271 compatible, they can always add an `isValidSignature` wrapper around this with the signature encoding trick you mentioned.

gakonst: I am comfortable with that.

src/Delegation.sol
```diff
@@ -582,10 +604,12 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
             keyHash = LibBytes.loadCalldata(signature, n);
             signature = LibBytes.truncatedCalldata(signature, n);
             // Do the prehash if last byte is non-zero.
+            // TODO: When do we use this?
             if (uint256(LibBytes.loadCalldata(signature, n + 1)) & 0xff != 0) {
                 digest = EfficientHashLib.sha2(digest); // `sha256(abi.encode(digest))`.
             }
```
gakonst, line 610: @Vectorized @jxom

src/Delegation.sol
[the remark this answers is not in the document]
gakonst: Would go extremely deep in the docs for this.

src/Delegation.sol
```diff
@@ -611,6 +635,29 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
...
+                mstore(add(m, 0x40), keyHash)
+                mstore(add(m, 0x60), 0x60) // signature offset
+                mstore(add(m, 0x80), signature.length) // signature length
+                calldatacopy(add(m, 0xa0), signature.offset, signature.length) // copy data to memory offset
+
+                let size := add(signature.length, 0x84)
+                let success := staticcall(gas(), signer, add(m, 0x1c), size, 0x00, 0x20)
+
+                // MagicValue: bytes4(keccak256("isValidSignatureWithKeyHash(bytes32,bytes32,bytes)")
+                if and(success, eq(shr(224, mload(0x00)), 0x8afc93b4)) { isValid := true }
```
gakonst, line 659: Why are these magic values even needed?

src/Delegation.sol
```diff
@@ -697,7 +749,11 @@ contract Delegation is IDelegation, EIP712, GuardedExecutor {
             computeDigest(calls, nonce), LibBytes.sliceCalldata(opData, 0x20)
         );
         if (!isValid) revert Unauthorized();
+
+        // TODO: Figure out where else to add these operations, after removing delegate call.
+        LibTStack.TStack(_KEYHASH_STACK_TRANSIENT_SLOT).push(keyHash);
         _execute(calls, keyHash);
+        LibTStack.TStack(_KEYHASH_STACK_TRANSIENT_SLOT).pop();
```
gakonst, line 756: ?

src/MultiSigSigner.sol
gakonst: I don't udnerstand this, extra docs pls

src/MultiSigSigner.sol
```diff
@@ -0,0 +1,198 @@
...
+
+            // This means that the keyHash was not found
+            if (j == config.ownerKeyHashes.length) {
+                return _FAIL_VALUE;
+            }
+        }
+
+        // If we reach here, then the required threshold was not met.
+        return _FAIL_VALUE;
+    }
```
gakonst, line 197: Can I re-use the same signature? I worked around that by sorting the signatures by signer and requiring the signer to be sorted, to avoid the dup signature hack. Can you add an explicit test for that

gakonst (approved):
I think I buy this so let's get it merged to be unblocked from prototyping

src/MultiSigSigner.sol
legion2002: I'll add a test. But we work around this by striking out the ownerKeyHash in the memory, so that it cannot be reused. (Line 174)

src/Delegation.sol
legion2002: This is coming from the 1271 standard. This costs no extra gas, but allows you to explicitly set true and false values, and different variations of true/false if needed.

PR #156, opened 2025-05-12
Update deployment scripts+docs

jenpaff:
This PR:
* Adds new toggle `SAVE_DEPLOY` to deploy scripts which writes deployed contract addresses to `deployment.json` when enabled. 
* Adds a new forge script which only deploys EIP7702Proxy+Delegation for account upgrades. 
* Updates deployment.md with docs on deployment.


deployments.json
```diff
@@ -0,0 +1 @@
+{"chainId":28404,"contracts":{"EntryPoint":"0x1029eD04911836Ab1019820c034a86D378275689","Delegation":"0x9aF7962fed8112Ba92Ce2B32D9DD557fCFcA6648","EIP7702Proxy":"0xD1C2Aa2E771F10C1ade71BCEd37a8201A6f37bf5","AccountRegistry":"0xdE6E47D09a12ca4C857EaB988aF4C554D78E6112","Simulator":"0x774D8B15621Aa58C97e7401b5ee001162b004054"}}
```
legion2002, line 1: We should also add the pauseAuthority address here.
In the current deployment setup, this is just set to the address of the deployer key.

jenpaff: added

klkvr: is this intended to be commited? if so, I think we should have it keyed by chainid so that we can maintain multiple deployments like this

jenpaff: yep thx for flagging, that was the intention. will fix and deploy new contracts on all our envs

legion2002 (approved):

PR #158, opened 2025-05-14
Catch Payment Errors in Simulate Loop

legion2002:
In the combinedGas loop, the payment is also incremented along with the combined gas.
There might come a point, where the user does not have enough funds for the payment. The user cannot recover from this, so we revert in the loop.

### DIFF
--- a/src/Simulator.sol
+++ b/src/Simulator.sol
@@ -177,7 +177,10 @@ contract Simulator {
                 if iszero(gasUsed) {
                     let m := mload(0x40)
                     returndatacopy(m, 0x00, returndatasize())
-                    revert(m, returndatasize())
+                    // `PaymentError` is given special treatment here, as it comes from
+                    // the account not having enough funds, and cannot be recovered from,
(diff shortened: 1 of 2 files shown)


src/Simulator.sol
```diff
@@ -172,6 +172,18 @@ contract Simulator {
...
 
+            // If the simulation failed, bubble up the full revert.
+            assembly ("memory-safe") {
+                if iszero(gasUsed) {
+                    let m := mload(0x40)
+                    returndatacopy(m, 0x00, returndatasize())
+                    // `PaymentError` is given special treatment here, as it comes from
+                    // the account not having enough funds, and cannot be recovered from,
+                    // since the paymentAmount will keep increasing in this loop.
+                    if eq(shr(224, mload(0x00)), 0xabab8fc9) { revert(m, 32) }
```
Vectorized, line 183: Nit:

```suggestion
                    if eq(shr(224, mload(0x00)), 0xabab8fc9) { revert(m, 0x20) }
```

PR #159, opened 2025-05-14
feat: forever period to GuardedExecutor

Vectorized:
**Use case:**

App asks user to authorize some session key that is authorize to pull a fixed amount of assets _only once_. 

The pulled assets could then go into some sub account where some bot could run automated trading on.

**Note:**

Due to the whitelist nature of our spend limits, a trading bot cannot work directly on the EOA using a P256 key. Say if it has swapped 1000 USDC to some memecoin, a separate spend permission has to be granted by a superAdmin to swap that memecoin out to sell. Granting superAdmins is a very risky move. The way around this is sub accounts.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,4 +4,3 @@
 [submodule "lib/solady"]
 	path = lib/solady
 	url = https://github.com/vectorized/solady
-	branch = main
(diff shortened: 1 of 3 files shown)


legion2002 (approved):

legion2002:
lgtm. Good PR to get us started to think about our subaccount design.

PR #167, opened 2025-05-19
ERC-1271 External KeyType

Vectorized:
For reference. 

May or may not merge.

Right now, the delegation does have some `msg.sender` dependent workflows on the ERC-1271 too.


legion2002:
I think it's desirable to have a non-standard interface for signers. To avoid confusion for integrators.
Discussed it [here](https://github.com/ithacaxyz/account/pull/153#discussion_r2096373028)

PR #191, opened 2025-06-03
Rename PortoAccount to IthacaAccount

legion2002:


legion2002:
> there are still some mentions to Porto in other non core files

Yeah I'll update those when we have a new name for the project.

PR #193, opened 2025-06-04
Fix the encoded intent arguments construction

lightning-li:


legion2002 (approved):

src/Orchestrator.sol
```diff
@@ -339,7 +339,7 @@ contract Orchestrator is
                 // and the sequence for `nonce` has been incremented.
                 // For more information, see `selfCallPayVerifyCall537021665()`.
                 selfCallSuccess :=
-                    call(g, address(), 0, add(m, 0x1c), add(encodedIntent.length, 0x44), 0x00, 0x20)
+                    call(g, address(), 0, add(m, 0x1c), add(encodedIntent.length, 0x24), 0x00, 0x20)
```
Vectorized, line 342: Optimization is correct.

- Selector (4)
- `simulationFlags` (32)
- `encodedIntent` (`encodedIntent.length`)

We have passed in an excess word that is unused.

Vectorized (approved):
This PR is correct, please merge it.

It fixes a hidden abi.encoding mistake that could break if the code is changed.

Thank you for the very sharp eye.

PR #197, opened 2025-06-10
Interop Design using Merkle Sigs

legion2002:
Interop design that adds merkle sigs natively into the orchestrator, allowing us to reuse the intent struct.

We also add another `fundTransfer` field to the `Intent` struct, which consists of encoded `<token,amount>` tuples that the relay _has_ to fund the user's account with before completing the output intent.

The output intent is defined as the intent that does the user's intended action on the destination chain.

The intput intents are used to pull funds from all the origin chains into a settlement system.

The Interop design is agnostic of what settlement system the relay uses.

Note: we only use the new murky library for tests, to create the merkle sigs. The actual source code only uses the solady merkle proof library.

### DIFF
--- a/test/Orchestrator.t.sol
+++ b/test/Orchestrator.t.sol
@@ -1364,6 +1364,50 @@ contract OrchestratorTest is BaseTest {
         assertEq(uint256(bytes32(errs[0])), 0);
         vm.assertEq(usdcMainnet.balanceOf(makeAddr("FRIEND")), 1000);
 
+        // 6. Attempt execution with duplicated or unordered `encodedFundTransfers` (should fail).
+        vm.revertToState(snapshot);
+        vm.chainId(1);
+        {
+            // Relay/Settlement system funds setup on Mainnet again.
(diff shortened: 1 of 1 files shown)


gakonst (commented):
good! let's do unhappy path tests!

Vectorized (approved):
lgtm. i’ll think of opts later if desired

Vectorized (changes requested):
Need to refactor `multichainExecute` such that each intent can be executed in a way that:
- Doesn't allocate new memory permanently for the remaining intents.
- Doesn't cause the rest of the intents to fail if one intent fails (such as due to failure to send funds).

legion2002:
> Need to refactor `multichainExecute` such that each intent can be executed in a way that:
> 
> * Doesn't allocate new memory permanently for the remaining intents.
> * Doesn't cause the rest of the intents to fail if one intent fails (such as due to failure to send funds).

Updated to an even cleaner version, that reuses the execute function.

PR #203, opened 2025-06-16
Adds language agnostic claude md 

legion2002:
Will add solidity specific instructions in a follow up PR after testing them out for some time.


insulineru (approved):
LFG! Very helpful part for future contributions, especially from guys who like AI coding. 

Thank you for bringing this to projects of this great caliber and making it the default in Web 3.0

PR #216, opened 2025-06-27
feat: LayerZeroSettler

Vectorized:
Todo: fix tests.

This merges into the LayerZeroSettler branch, not main.

### DIFF
--- a/src/LayerZeroSettler.sol
+++ b/src/LayerZeroSettler.sol
@@ -65,10 +65,6 @@ contract LayerZeroSettler is OApp, ISettler {
                 dstEid, payload, allOptions[i], MessagingFee(nativeFees[i], 0), payable(msg.sender)
             );
         }
-
-        // if (msg.value > totalFee) {
-        //     TokenTransferLib.safeTransfer(address(0), msg.sender, msg.value - totalFee);
-        // }
     }
 
     /// @notice Receive settlement attestation from another chain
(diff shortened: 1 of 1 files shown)


src/LayerZeroSettler.sol
```diff
@@ -26,26 +30,40 @@ contract LayerZeroSettler is OApp, ISettler {
     /// @dev Requires msg.value to cover all LayerZero fees
     function send(bytes32 settlementId, bytes calldata settlerContext) external payable override {
         // Decode settlerContext as an array of LayerZero endpoint IDs
-        uint32[] memory endpointIds = abi.decode(settlerContext, (uint32[]));
+        (uint128[] memory expectedDstGasRequired, uint32[] memory endpointIds) =
```
legion2002, line 33: Why do we need this gas stuff?
We only need it if we are using executors from layer zero I think.

But we are going to make the execution ourselves using the relay. So we only use l0 DVNs

Vectorized: ok, if we are going to DIY the execution with the relay, then maybe it is ok?

i need to check.

legion2002:
@Vectorized Did you have any thoughts on this PR? If we are running the relay as an executor, then do we still need these changes?

Vectorized:
I'll close this for now. If we need it, then we'll copy the code over.

PR #228, opened 2025-07-03
feat: allow withdrawals via owner signature

klkvr:
Adds support for withdrawals via an EIP-712 signature from owner EOA. That way we're not forced to have direct access to owner private key to perform rebalances

### DIFF
--- a/src/SimpleFunder.sol
+++ b/src/SimpleFunder.sol
@@ -25,7 +25,7 @@ contract SimpleFunder is EIP712, Ownable, IFunder {
     error InvalidFunderSignature();
     error InvalidWithdrawalSignature();
     error InvalidNonce();
-    error Expired();
+    error DeadlineExpired();
 
     address public immutable ORCHESTRATOR;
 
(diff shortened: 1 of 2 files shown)


src/SimpleFunder.sol
```diff
@@ -48,6 +68,32 @@ contract SimpleFunder is Ownable, IFunder {
...
 
+    /// @dev Allows to withdraw tokens via a signature from owner.
+    function withdrawTokensWithSignature(
+        address token,
+        address recipient,
+        uint256 amount,
+        uint256 deadline,
+        uint256 nonce,
+        bytes calldata signature
+    ) external onlyOwner {
```
legion2002, line 79: I think we want to remove the onlyOwner here

src/SimpleFunder.sol
```diff
@@ -48,6 +68,32 @@ contract SimpleFunder is Ownable, IFunder {
         TokenTransferLib.safeTransfer(token, recipient, amount);
     }
 
+    /// @dev Allows to withdraw tokens via a signature from owner.
+    function withdrawTokensWithSignature(
+        address token,
+        address recipient,
+        uint256 amount,
+        uint256 deadline,
```
legion2002, line 76: This deadline is not validated anywhere.

klkvr: fixed in dd33336

PR #231, opened 2025-07-04
chore: multichain infra deployments

legion2002:


script/DeployAll.s.sol
```diff
@@ -27,6 +31,8 @@ contract DeployAllScript is Script {
                 vm.envAddress("FUNDER"), address(orchestrator), vm.envAddress("FUNDER_OWNER")
             )
         );
+        simpleSettler = address(new SimpleSettler(vm.envAddress("SETTLER_OWNER")));
```
jenpaff, line 34: Just for context: what's the role of the settler owner?

jenpaff (approved):

script/DeployAll.s.sol
legion2002: The simple settler contract uses a trusted cross chain settlement model. The settler owner can do arbitrary cross chain settlement writes to any chains.

For testing, the relay would use this instead of the L0settler, as it is much simpler to test using this.

PR #247, opened 2025-07-15
fix: prePayment > totalPayment edge case

legion2002:


src/Orchestrator.sol
```diff
@@ -296,8 +296,11 @@ contract Orchestrator is
             LibBit.or(
                 i.prePaymentAmount > i.prePaymentMaxAmount,
                 LibBit.or(
-                    i.prePaymentMaxAmount > i.totalPaymentMaxAmount,
-                    i.totalPaymentAmount > i.totalPaymentMaxAmount
+                    i.totalPaymentAmount > i.totalPaymentMaxAmount,
+                    LibBit.or(
+                        i.prePaymentMaxAmount > i.totalPaymentMaxAmount,
+                        i.prePaymentAmount > i.totalPaymentAmount
```
Vectorized, line 302: https://github.com/Vectorized/solady/blob/a096f4fb0f65d1c6d6677ea6b13e9d41cb0bf798/src/utils/LibBit.sol#L263 use this 4-arg `or` variant for nicer looking code. I've added it for this very use case.

Vectorized (approved):

src/Orchestrator.sol
legion2002: @claude make a pr that merges into this PR, which uses this 4 argument or.

PR #251, opened 2025-07-16
refactor: move settler.send from orchestrator to account execution

legion2002:
## Summary

This PR refactors the settlement flow to move the `settler.send` call from the orchestrator to the account's execution data. This change enables accounts to directly trigger settlement attestation during output intent execution, removing the dependency on the orchestrator for this critical step.

### 1. Removed Native Settler Support from Orchestrator
- The orchestrator no longer makes direct calls to `settler.send`
- Settlement attestation is now the responsibility of the account executing the output intent

### 2. Updated Settlement ID Calculation Convention
- **Breaking the circular dependency**: The settlement ID is now calculated as the digest of the intent WITHOUT the `settler.send` call
- This resolves the circular dependency where the settlement ID is needed for the `settler.send` call, but if included in the intent, the digest would depend on itself

### 3. New Recommended Flow
1. Create the output intent with execution data containing only the actual output operations (e.g., token transfers)
2. Calculate the settlement ID as the digest of this intent
3. Update the intent's execution data to include the `settler.send` call with the calculated settlement ID

## Important Security Consideration: Settlement ID Uniqueness

**⚠️ Critical**: It is extremely important to choose the settlement ID correctly. If you use a settlement ID that has already been used in a previous settlement, you risk having escrows resolve according to the old settlement attestation.

This could lead to:
- Incorrect fund releases from escrows
- Settlement mismatches between chains
- Potential loss of funds if old attestations are reused

The recommended convention (using the intent digest without the settler call) helps ensure uniqueness because:
- Each intent has a unique nonce
- The execution data represents the unique set of operations
- The combination creates a unique identifier for each settlement

## Documentation Updates

Added comprehensive documentation to `ISettler.send` function explaining the "Convention for choosing settlement ID" to guide future implementations and prevent security issues.

## Test Updates

- Updated `testMultiChainIntent` to follow the new pattern
- Changed expectations to account for `settler.send` being called by the account instead of the orchestrator
- Updated all `settler.write` calls to use the account address as the sender

## Migration Guide

For existing implementations:
1. Remove any direct `settler.send` calls from orchestrator interactions
2. Include the `settler.send` call in your output intent's execution data
3. Calculate settlement ID using the intent digest (without the settler call) to ensure uniqueness
4. Ensure your escrow contracts check the sender correctly (now the account, not orchestrator)

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/IthacaAccount.sol
+++ b/src/IthacaAccount.sol
@@ -726,6 +726,6 @@ contract IthacaAccount is IIthacaAccount, EIP712, GuardedExecutor {
         returns (string memory name, string memory version)
     {
         name = "IthacaAccount";
-        version = "0.4.8";
+        version = "0.4.9";
     }
 }
--- a/src/Orchestrator.sol
+++ b/src/Orchestrator.sol
@@ -889,7 +889,7 @@ contract Orchestrator is
         returns (string memory name, string memory version)
     {
         name = "Orchestrator";
-        version = "0.4.5";
+        version = "0.4.6";
     }
 
     ////////////////////////////////////////////////////////////////////////


legion2002:
Closing this PR, because either settler.send() and funder.fund() need to be atomic. If we move settler.send() to the execution batch, and the execution reverts for some reason, we also want the funding to revert, otherwise the user basically gets free money.

An alternative is that we also move funder.fund() to the execution batch, this seems cleaner on the top level, but requires bigger changes for how the funding works. It also adds security concerns about hwo the user ensures that the funder is being honest ( because this needs to be done onchain )

Current recommendation is to enshrine everything in the orchestrator as a mediator, we should think very carefully about the security implications of negative cases before moving things into the execution data.

PR #257, opened 2025-07-21
fix: Zeroize scratch space for return data

Vectorized:


legion2002 (commented):
Does this PR have the latest solady version? Or is this regressing to an older one?

Issue #152, opened 2025-05-09
Separate Simulation Contracts from Deployed Contracts


legion2002:
Although it is very simple to reason about when a simulation flag is set to true, it is another critical security piece that we always have  to keep in mind.
If we somehow mess up the simulation flag somewhere, then effectively an attacker can have complete control over all user accounts.
I think a much safer alternative is that we remove all simulation related logic from the contracts that will be deployed. And we maintain a separate Delegation contract called the SimulateDelegation in the repo.

The real Delegation & the SimulateDelegation will only have a 1 line diff. In the SimulateDelegation, the unwrapAndValidateSignature function will always return true. ( We can make all this look nice, and be maintainable, using function overrides )

All the relay has to do, is before simulating a userOp, they also have to state override the code of the Delegation contract, with the SimulateDelegation bytecode.

PR #258, opened 2025-07-22
Simplification & Fixes in Orchestrator and Account

legion2002:
1. Solves issues mentioned in the audit around misuse of combinedGas field, and minimum gas needed by the execute function. 
2. Removes all simulation logic from deployed contracts, making core contracts even simpler, without any loss of accuracy. #152 
3. Extracts batching of intents from the core contracts to the periphery. Which allows us to remove a lot of ugly logic around handling errors. The current way gives a lot of flexibility to the relay to batch intents how they want. The resulting errors are also more detailed, and should lead to easier debugging.

Note all of the changes, come at no regression in gas costs or features, but lead to reducing a lot of tech debt, and solve a bunch of audit issues.

### DIFF
--- a/src/Orchestrator.sol
+++ b/src/Orchestrator.sol
@@ -73,6 +73,8 @@ contract Orchestrator is
     /// @dev Out of gas to perform the call operation.
     error InsufficientGas();
 
+    error InsufficientAccountExecuteGas();
+
     /// @dev The order has already been filled.
     error OrderAlreadyFilled();
 
@@ -105,7 +107,7 @@ contract Orchestrator is
     /// This event is emitted in the `execute` function.
     /// - `nonce` denotes the nonce that has been incremented to invalidate `nonce`.
     /// For PreCalls where the nonce is skipped, this event will NOT be emitted.
-    event IntentExecuted(address indexed eoa, uint256 indexed nonce);
+    event IntentExecuted(address indexed eoa, uint256 indexed nonce, bool success);
 
     ////////////////////////////////////////////////////////////////////////
     // Constants
@@ -195,10 +197,16 @@ contract Orchestrator is
     }
 
     /// @dev Executes a single encoded intent.
-    // TODO: do we need something like the combinedGasOverride anymore?
-    function execute(bytes calldata encodedIntent) public payable virtual nonReentrant {
+    function execute(bytes calldata encodedIntent)
+        public
+        payable
+        virtual
+        nonReentrant
+        returns (bytes memory err)
+    {
         Intent calldata i = _extractIntent(encodedIntent);
 
+        _checkCallDepth(i.accountExecuteGas);
         if (
             LibBit.or(
                 i.prePaymentAmount > i.prePaymentMaxAmount,
@@ -290,7 +298,22 @@ contract Orchestrator is
         // off-chain simulation and on-chain execution.
         if (i.prePaymentAmount != 0) _pay(i.prePaymentAmount, keyHash, digest, i);
 
-        console.log("B EOA: ", i.eoa);
+        try this.selfCallTest(keyHash, digest, encodedIntent) {
+            emit IntentExecuted(i.eoa, i.nonce, true);
+            err = "";
+        } catch {
+            // console.log("Inside catch");
+            emit IntentExecuted(i.eoa, i.nonce, false);
+
+            assembly ("memory-safe") {
+                let size := returndatasize()
+                mstore(err, size)
+                returndatacopy(add(err, 0x20), 0x00, size)
+            }
+            // console.log("size", size);
+        }
+
+        // _selfCall(keyHash, digest, encodedIntent);
 
         // Equivalent Solidity code:
         // try this.selfCallExecutePay( keyHash, i) {}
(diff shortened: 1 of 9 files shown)


src/IthacaAccount.sol
```diff
@@ -627,13 +627,6 @@ contract IthacaAccount is IIthacaAccount, EIP712, GuardedExecutor {
             // Set the target key hash to the payer's.
             keyHash = k;
```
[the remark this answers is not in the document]
Vectorized: i like this removal. now the simulation functionality is more isolated.

src/Batcher.sol
```diff
@@ -0,0 +1,44 @@
...
+        if (length != intentGas.length) revert ArrayLengthMismatch();
+
+        // Execute each intent
+        for (uint256 i = 0; i < length; ++i) {
+            // Encode the function call
+            bytes memory data =
+                abi.encodeWithSelector(IOrchestrator.execute.selector, encodedIntents[i]);
+
+            // Make the call with specified gas
+            (bool success, bytes memory returnData) = orchestrator.call{gas: intentGas[i]}(data);
```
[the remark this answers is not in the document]
Vectorized: This comment is invalid.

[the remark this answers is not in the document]
Vectorized: Personally, I would write this from assembly from the ground up, but for clarity, this implementation thus far is good.

src/Orchestrator.sol
```diff
@@ -21,6 +21,8 @@ import {IFunder} from "./interfaces/IFunder.sol";
 import {ISettler} from "./interfaces/ISettler.sol";
 import {MerkleProofLib} from "solady/utils/MerkleProofLib.sol";
 
+import "forge-std/console.sol";
```
howydev, line 24:
```suggestion

```

src/Orchestrator.sol
```diff
@@ -481,8 +290,10 @@ contract Orchestrator is
         // off-chain simulation and on-chain execution.
         if (i.prePaymentAmount != 0) _pay(i.prePaymentAmount, keyHash, digest, i);
 
+        console.log("B EOA: ", i.eoa);
+
         // Equivalent Solidity code:
-        // try this.selfCallExecutePay(flags, keyHash, i) {}
+        // try this.selfCallExecutePay( keyHash, i) {}
```
howydev, line 319: nit 

```suggestion
        // try this.selfCallExecutePay(keyHash, i) {}
```

src/Orchestrator.sol
```diff
@@ -495,52 +306,65 @@ contract Orchestrator is
             let m := mload(0x40) // Load the free memory pointer
             mstore(0x00, 0) // Zeroize the return slot.
             mstore(m, 0x00000001) // `selfCallExecutePay1395256087()`
-            mstore(add(m, 0x20), flags) // Add flags as first param
-            mstore(add(m, 0x40), keyHash) // Add keyHash as second param
-            mstore(add(m, 0x60), digest) // Add digest as third param
+            mstore(add(m, 0x20), keyHash) // Add keyHash as second param
+            mstore(add(m, 0x40), digest) // Add digest as third param
+            mstore(add(m, 0x60), 0x80) // Add offset of encoded Intent as third param
+            let encodedIntentLength := sub(calldatasize(), 0x44)
```
howydev, line 371: this is potentially dangerous i think, a valid calldata for the top level execute might be:

`fnSelector | offset | junk data | length | intent`

then this would end up grabbing `junk data`

src/Orchestrator.sol
```diff
@@ -495,52 +306,65 @@ contract Orchestrator is
...
-                    revert(mload(0x40), returndatasize())
-                }
-                return(m, 0x20)
+                // TODO: We should not revert here.
+                // returndatacopy(mload(0x40), 0x00, returndatasize())
+                // revert(mload(0x40), returndatasize())
             }
         }
+
+        emit IntentExecuted(i.eoa, i.nonce);
```
howydev: have a slight preference to keep the success/failure but conditionally emit this, otherwise it might not provide very useful data if it also contains execution failures

src/Orchestrator.sol
```diff
@@ -495,52 +306,65 @@ contract Orchestrator is
...
-            i := add(0x64, calldataload(0x64))
+            /*
+            calldata:
+            0x00: 0x00000001
+            0x04: keyHash
+            0x24: digest
+            0x44: 0x80 ( offset of encodedIntent)
+            0x64: encodedIntentLength
+            0x84: encodedIntent data
+                0x84: 0x20 ( offset of intent struct data, from starting of encodedIntent)
```
howydev, line 433: same as above, potentially risky to rely on this as the encoding

src/Orchestrator.sol
```diff
@@ -613,19 +440,14 @@ contract Orchestrator is
...
-                    revert(0x00, 0x20) // Revert the `err` (NOT return).
+                    returndatacopy(mload(0x40), 0x00, returndatasize())
+                    revert(mload(0x40), returndatasize())
                 }
             }
 
             // Event so that indexers can know that the nonce is used.
             // Reaching here means there's no error in the PreCall.
-            emit IntentExecuted(eoa, p.nonce, true, 0); // `incremented = true`, `err = 0`.
+            emit IntentExecuted(eoa, p.nonce);
```
howydev: should we differentiate between the precall events and the main intent events?

src/simulation/OrchestratorSim.sol
```diff
@@ -0,0 +1,71 @@
...
+    /// @return gasUsed The amount of gas used by the execution. (Only returned if `isRevert` is true)
+    /// gAccountExecute is the amount of gas used by the account.execute function.
+    /// (Includes the gas cost, to make the call)
+    function simulateExecute(bool isRevert, bytes calldata encodedIntent)
+        external
+        payable
+        returns (uint256 /*gAccountExecute*/ )
+    {
+        // If Simulation fails, then it will revert here.
+        execute(encodedIntent);
```
howydev, line 39: iiuc this would be an internal call, so all the `calldataX` operations operate on the top level calldata that would include `isRevert`

PR #267, opened 2025-07-31
fix: switch from using balance delta to determine if a {ERC20, native} token transfer succeeded

howydev:
Change payments to be a pull rather than a push for `pay` and `fund` to be able to differentiate between account griefing the payment relayer griefing the payment receipt. In the latter case, for post execution payments, since verification passes the account is still charged for the intent 

gas cost for native and erc20 token flows goes up by ~2k

### DIFF
--- a/src/SimpleFunder.sol
+++ b/src/SimpleFunder.sol
@@ -158,7 +158,7 @@ contract SimpleFunder is EIP712, Ownable, IFunder {
                 // Orchestrator checks for token transfer success, so we don't need to check it here.
                 pop(call(gas(), token, 0, add(m, 0x10), 0x44, 0, 0x20))
 
-                let allowance := mload(0x20)
+                let allowance := mload(0)
                 if gt(amount, allowance) {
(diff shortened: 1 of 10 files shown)


src/SimpleFunder.sol
```diff
@@ -139,8 +137,36 @@ contract SimpleFunder is EIP712, Ownable, IFunder {
...
+        }
+
+        address orchestrator = ORCHESTRATOR;
+        for (i; i < transfers.length; ++i) {
+            address token = transfers[i].token;
+            uint256 amount = transfers[i].amount;
+            // We check if the token has already been approved to the orchestrator. If not, do a max approval.
+            assembly ("memory-safe") {
+                let m := mload(0x40)
+                mstore(add(m, 0x40), orchestrator)
```
legion2002, line 155: nit: assembly mstores should be in asc order m, add(m,0x20), add(m,0x40)

legion2002 (approved):

PR #269, opened 2025-08-04
feat: move simulation check to use `tx.origin`

howydev:
Modifying the simulation condition check to `tx.origin.balance >= uint192max` instead of `address(oc).balance == uint256max`

This unblocks improving trust assumptions of payment flows to route through the orchestrator instead of from eoa/account directly to the relayer

### DIFF
--- a/src/interfaces/IOrchestrator.sol
+++ b/src/interfaces/IOrchestrator.sol
@@ -25,7 +25,7 @@ interface IOrchestrator is ICommon {
     /// But the codepaths for signature verification are still hit, for correct gas measurement.
     /// @dev If `isStateOverride` is false, then this function will always revert. If the simulation is successful, then it reverts with `SimulationPassed` error.
     /// If `isStateOverride` is true, then this function will not revert if the simulation is successful.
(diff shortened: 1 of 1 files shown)


howydev:
1. i updated the address to use 0xAa239C49EC6D564597F2e2F99A357c63cb65090d which is keccak256("Ithaca.Orchestrator.SIMULATION")[12]. happy to change this to any other value too
2. i changed the SimpleFunder test to dynamically get the version, else tests will fail on CI version bump

src/interfaces/IOrchestrator.sol
```diff
@@ -25,7 +25,7 @@ interface IOrchestrator is ICommon {
     /// But the codepaths for signature verification are still hit, for correct gas measurement.
     /// @dev If `isStateOverride` is false, then this function will always revert. If the simulation is successful, then it reverts with `SimulationPassed` error.
     /// If `isStateOverride` is true, then this function will not revert if the simulation is successful.
-    /// But the balance of msg.sender has to be equal to type(uint256).max, to prove that a state override has been made offchain,
+    /// But the balance of address(0) has to be equal to type(uint256).max, to prove that a state override has been made offchain,
```
legion2002, line 28: outdated comment

PR #274, opened 2025-08-05
feat: add create2 deployments

legion2002:


deploy/DeployMain.s.sol
```diff
@@ -19,6 +19,12 @@ import {LayerZeroSettler} from "../src/LayerZeroSettler.sol";
  * @notice Main deployment script that executes all configured stages for specified chains
  * @dev This script directly deploys contracts without creating intermediate deployer contracts
  *
+ * ⚠️  IMPORTANT FOR CREATE2 DEPLOYMENTS:
+ * If you're using CREATE2 (salt != 0x0), you MUST save your salt values!
+ * The salt determines your contract addresses across chains.
+ * Lost salts = unable to deploy to same addresses on new chains.
+ * Store production salts securely with backups.
```
jenpaff, line 26: as in store them in 1password?

deploy/DefaultConfig.sol
```diff
@@ -157,6 +166,7 @@ contract DefaultConfig {
             l0SettlerOwner: 0x70997970C51812dc3A010C7d01b50e0d17dc79C8,
             layerZeroEndpoint: 0x0000000000000000000000000000000000000000,
             layerZeroEid: 0,
+            salt: bytes32(0),
```
jenpaff, line 169: why do we have one entry per chain if it should be the same across all chains?

PR #278, opened 2025-08-07
fix: move approvals ordering

howydev:


legion2002 (approved):
good catch, lgtm! 

But would like @Vectorized to also look and think about more deeply, before we merge it.

PR #280, opened 2025-08-11
refactor(LibTStack): declare EmptyStack error for proper revert decoding

Galoretka:
Declare error EmptyStack(); at file scope in src/libraries/LibTStack.sol
Keeps existing assembly reverts (selector 0xbc7ec779) — no behavior change
Improves readability and enables tooling/decoders to resolve the revert
Affects top and pop paths only


legion2002:
Thanks

legion2002 (approved):

PR #282, opened 2025-08-11
fix: 1271 replay issue

howydev:
add a simple replay safe wrapper

### DIFF
--- a/src/IthacaAccount.sol
+++ b/src/IthacaAccount.sol
@@ -186,6 +186,9 @@ contract IthacaAccount is IIthacaAccount, EIP712, GuardedExecutor {
     /// @dev For EIP712 signature digest calculation for the `execute` function.
     bytes32 public constant CALL_TYPEHASH = keccak256("Call(address to,uint256 value,bytes data)");
 
+    /// @dev For ERC1271 replay-safe hashing.
+    bytes32 public constant SIGN_TYPEHASH = keccak256("Sign(bytes32 digest)");
+
     /// @dev For EIP712 signature digest calculation.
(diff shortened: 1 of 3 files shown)


src/IthacaAccount.sol
```diff
@@ -233,6 +236,9 @@ contract IthacaAccount is IIthacaAccount, EIP712, GuardedExecutor {
         virtual
         returns (bytes4)
     {
```
legion2002, line 238: Can you add some natspec for integrators, that explains how to sign signatures for 1271.

howydev: i added it in the code block since natspec was already pretty fat

legion2002 (approved):

src/IthacaAccount.sol
```diff
@@ -233,6 +236,13 @@ contract IthacaAccount is IIthacaAccount, EIP712, GuardedExecutor {
...
+        // To sign an app digest (e.g. Permit2), you would need to perform a `hashTypedData` on the app's 712,
+        // along with the app's domain, then `signTypedData` with the account's 712 and account domain.
+        // The account domain is added as a layer to prevent replay attacks since some apps do not include the
+        // account address as a field in their 712 data.
+        bytes32 replaySafeDigest = EfficientHashLib.hash(SIGN_TYPEHASH, digest);
+        digest = _hashTypedData(replaySafeDigest);
+
```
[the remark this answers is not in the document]
howydev: r1 signatures are replay safe? is it because each passkey will always only be tied to a single account?

PR #295, opened 2025-08-14
Tanishk/toml configs

legion2002:


foundry.toml
```diff
@@ -4,13 +4,17 @@
...
+extends = "deploy/config.toml"
 evm_version = "prague" # Cancun will be tested in the CI.
 auto_detect_solc = false
 optimizer = true
 optimizer_runs = 200
 gas_limit = 100_000_000 # ETH is 30M, but we use a higher value.
 fs_permissions = [
-    { access = "read-write", path = "deploy/registry" }
+    { access = "read-write", path = "deploy/registry" },
+    { access = "read", path = "deploy/config.toml" }
```
0xrusowsky, line 17: note that u don't need ffi permissions as u are already signaling that we should read from this file with `extends`

jenpaff (approved):

foundry.toml
legion2002: Tried this, doesn't work without ffi permission

PR #304, opened 2025-08-21
fix: add calldata offset checks for safety

howydev:


src/Orchestrator.sol
```diff
@@ -252,10 +255,21 @@ contract Orchestrator is
...
+
+            // Just in case, we do bounds check on the offsets of the struct and the inner fields.
+            // We sum up the offsets after doing 2**64 shift and expect them to be == 0.
+            let s := add(shr(64, t), shr(64, calldataload(add(i, 0x20)))) // executionData
+            s := add(s, shr(64, calldataload(add(i, 0xe0)))) // bytes[] encodedPreCalls
+            s := add(s, shr(64, calldataload(add(i, 0x100)))) // bytes[] encodedFundTransfers
+            s := add(s, shr(64, calldataload(add(i, 0x1a0)))) // bytes funderSignature
+            s := add(s, shr(64, calldataload(add(i, 0x1c0)))) // bytes settlerContext
+            s := add(s, shr(64, calldataload(add(i, 0x220)))) // bytes signature
+            s := add(s, shr(64, calldataload(add(i, 0x240)))) // bytes paymentSignature
```
Vectorized, line 267: `or` feels better, but lgtm.

Vectorized (approved):

Vectorized:
Ok, the hard part is... can we make a script to auto generate all these?

Vectorized:
@howydev Merge #310 first. Then we can avoid updating all these constants in the future.

If we _really really_ want to patch the relay calldata griefing vector (tho users can switch relays if one relay is fleecing them), we can simply restrict `encodedIntent.length` to be under a reasonable size (e.g. 256kb, which constitutes around 100k worth of memory expansion gas, sufficient for contract deployments via factories, post quantum sigs, etc).

legion2002 (approved):

PR #307, opened 2025-08-22
feat: add ExpToken deployment to DeployMain script

legion2002:
## Summary

Cherry-picked commit that adds ExpToken (EXP and EXP2) deployment functionality to the DeployMain script for testnet chains.

### DeployMain Script
- Added ExpToken deployment function that creates two tokens: "EXP" and "EXP2"
- Testnet-only deployment logic - ExpToken only deploys when `is_testnet = true`
- Integration with "ALL" contracts filter - ExpToken automatically included for testnets
- Required configuration fields: `exp_minter_address` and `exp_mint_amount`
- Both tokens mint to the same address with the same amount
- CREATE2 deployment for deterministic addresses
- Registry tracking for both token addresses

### Configuration
- Added EXP token configuration to all testnet chains in config.toml
- Required fields with no defaults - deployment fails if missing on testnets

### Documentation
- Updated deploy/README.md with comprehensive ExpToken documentation
- Removed old DeployEXP script references
- Added configuration examples and behavior explanation
- Registry file format documentation

## Testing

Verified deployment behavior:
- ✅ Testnets (`is_testnet = true`): Deploys 8 core contracts + ExpToken (9 total)
- ✅ Production (`is_testnet = false`): Deploys 8 core contracts only
- ✅ Both EXP and EXP2 tokens deployed with correct names and addresses
- ✅ Required configuration enforced - fails with clear error if missing

### DIFF
--- a/deploy/DeployMain.s.sol
+++ b/deploy/DeployMain.s.sol
@@ -15,6 +15,7 @@ import {SimpleFunder} from "../src/SimpleFunder.sol";
 import {Escrow} from "../src/Escrow.sol";
 import {SimpleSettler} from "../src/SimpleSettler.sol";
 import {LayerZeroSettler} from "../src/LayerZeroSettler.sol";
+import {ExperimentERC20} from "./mock/ExperimentalERC20.sol";
 
 /**
  * @title DeployMain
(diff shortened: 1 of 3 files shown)


legion2002:
@kuyziss 

> can we actually have it send funds to multiple addresses? bc we'd usually want to send funds to funder + the address we run stress tests from

We'll do this later, because it will break compatibility for the infra scripts we have for 0.4.6 
So waiting for everyone to move to 0.5.0, before making any changes. Is this okay?

howydev (approved):

PR #310, opened 2025-08-25
fix: bump solady, use checkInCalldata

Vectorized:
I don't really like hardcoded constants, just in case we forget to change them in the future.


src/Orchestrator.sol
```diff
@@ -252,25 +249,21 @@ contract Orchestrator is
...
         }
+        // These checks are included for more safety: Swiss Cheese Model.
+        // Ensures that all the dynamic children in `encodedIntent` are contained.
+        LibBytes.checkInCalldata(i.executionData, intentCalldata);
+        LibBytes.checkInCalldata(i.encodedPreCalls, intentCalldata);
+        LibBytes.checkInCalldata(i.encodedFundTransfers, intentCalldata);
+        LibBytes.checkInCalldata(i.funderSignature, intentCalldata);
+        LibBytes.checkInCalldata(i.settlerContext, intentCalldata);
+        LibBytes.checkInCalldata(i.signature, intentCalldata);
+        LibBytes.checkInCalldata(i.paymentSignature, intentCalldata);
```
Vectorized, line 266: Taste: I feel this could be `require(LibBytes.isInCalldata(child, parent)`. 

This will be more performant.

Vectorized: Nevermind, let's leave it as `checkInCalldata`, more semantically correct, since it abuses out-of-gas for the array case to revert.

snapshots/BenchmarkTest.json
```diff
@@ -1,43 +1,43 @@
...
-  "testNativeTransfer_IthacaAccount_ERC20SelfPay": "156721",
-  "testNativeTransfer_ZerodevKernel": "208623",
-  "testUniswapV2Swap_AlchemyModularAccount": "238779",
-  "testUniswapV2Swap_CoinbaseSmartWallet": "237559",
-  "testUniswapV2Swap_ERC4337MinimalAccount": "231242",
-  "testUniswapV2Swap_IthacaAccount": "189700",
-  "testUniswapV2Swap_IthacaAccount_AppSponsor": "206595",
-  "testUniswapV2Swap_IthacaAccount_ERC20SelfPay": "212153",
-  "testUniswapV2Swap_ZerodevKernel": "266487"
+  "testERC20Transfer_AlchemyModularAccount": "159052",
```
howydev, line 2: nit - these are the test values, we should run `forge snapshot --isolate` before each PR now

howydev (approved):

Issue #212, opened 2025-06-20
Go through all manual decoding in the contracts and simplify or think about attack vectors.


legion2002:
We currently use assembly to skip some bound checks and manually decode important structs like `Intent`, abi.decode adds some extra checks by default, we need to go through all the structs we decode in this way and make sure that we are handling them safely.

Issue #201, opened 2025-06-15
Add calldata manipulation tests for external structs


legion2002:
Currently for some critical structs like the `Intent` struct, we do some optimizations that skip solidity type checks explicitly, since we assume that these fields will be sanitized when solidity uses them throughout the code.

We need to add tests mainpulating each of these fields in the calldata, to make sure that manipulating these fields cannot affect the rest of the struct.

smitrajput:
hi @legion2002, noticed this is self-assigned. If help’s welcome, I can add a tests-only suite for calldata manipulation on the `Intent` struct (and others you name), no logic changes, with gas snapshots. I'll share a concrete approach once I get a green from you.

Useful? If not, would love to know if there's something more pressing you could use my help on.

legion2002:
Hi @smitrajput, please go ahead.

smitrajput:
Just noticed that the current call flow to `IthacaAccount.pay()` goes through all those functions (except `simulateExecute()`) in one go, so won't need separate tests for each, but one good test with intelligent corruption should be enough to reveal any vulnerabilities during unsafe decoding

<img width="668" height="324" alt="Image" src="https://github.com/user-attachments/assets/25acdeaf-6fbe-4ee1-b619-ce7c58797783" />

PR #312, opened 2025-08-25
test: Intent's static + dynamic offsets' field corruptions

smitrajput:
### Test Description
This is a good first step for issues [201](https://github.com/ithacaxyz/account/issues/201) and [212](https://github.com/ithacaxyz/account/issues/212), testing potentially unsafe decoding by corrupting the 13 static fields and offsets of all 7 dynamic fields of `Intent` struct, along with Intent's main offset, during calls to `Orchestrator.execute()`, targeting naked decoding at [`_extractIntent()`](https://github.com/ithacaxyz/account/blob/main/src/Orchestrator.sol#L244), [`selfCallPayVerifyCall537021665()`](https://github.com/ithacaxyz/account/blob/main/src/Orchestrator.sol#L411), [`_extractPreCall()`](https://github.com/ithacaxyz/account/blob/main/src/Orchestrator.sol#L270) and [`IthacaAccount.pay()`](https://github.com/ithacaxyz/account/blob/main/src/IthacaAccount.sol#L617). <br/><br/>Tests are split into 21 functions: 13 for corrupting static fields, 7 for dynamic field offsets and 1 corrupting the main intent offset.

### Result
All tests pass clean without the corruption (base case in existing tests), and throw `EvmError: Revert` after corruption for 6 dynamic field offset corruptions as **expected**, caught as `false` value returned on doing a [`.call() to Orchestrator.execute()`](https://github.com/ithacaxyz/account/pull/312/files#diff-ff27a0826a73dea2fc4faf0dc2536085954313b31363b75b6db3a4a66d71ad99R521). The remaining 15 return well-defined errors in the codebase which are caught neatly. Implying corrupted offsets are unable to induce unintended behaviours and the decoding throughout the call from `Orchestrator.execute()` to `IthacaAccount.pay()` is safe from this attack.

### Potential Improvement
Would have preferred a simpler multichain intent setup, to test the corrupted `settlerContext` field than the [current one](https://github.com/ithacaxyz/account/pull/312/files#diff-ff27a0826a73dea2fc4faf0dc2536085954313b31363b75b6db3a4a66d71ad99R793), which reuses most of `Orchestrator.t.sol`'s [testMultichainIntent()](https://github.com/ithacaxyz/account/blob/main/test/Orchestrator.t.sol#L1301) setup. Improvement suggestions welcomed.

### Future Work
Note that tests for corrupting the data of dynamic fields still remain. Worth adding in a future PR.

### DIFF
--- a/test/Account.t.sol
+++ b/test/Account.t.sol
@@ -421,6 +421,8 @@ contract AccountTest is BaseTest {
      *
      */
 
+    //////// Corrupting the 13 static fields of Intent ////////
+
     // Test 1: eoa corruption
     function testPayWithAllCorruptedEOAFieldOfIntent() public {
         bytes memory maliciousCalldata = _createIntentOnMainnet();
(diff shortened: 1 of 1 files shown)


smitrajput:
@legion2002 @howydev plz take a look and lemme know of improvements

legion2002:
@smitrajput Thanks we'll look into it soon, these tests will be very helpful!

test/Account.t.sol
```diff
@@ -498,6 +510,458 @@ contract AccountTest is BaseTest {
...
+
+        console.log("Test 2: executionData offset corruption");
+        maliciousCalldata = _createIntentOnMainnet();
+        assembly {
+            let dataPtr := add(maliciousCalldata, 0x20) // Skip bytes length prefix
+            let intentPtr := add(dataPtr, 0x20) // Points to start of Intent struct
+            // executionData offset (bytes 64-95 relative to start, or 32-63 in Intent struct)
+            mstore(add(intentPtr, 32), 0x10000000000000001) // 2^64 + 1
+        }
+        // assertEq(oc.execute(maliciousCalldata), bytes4(keccak256("VerifiedCallError()")));
```
howydev, line 536: nit: remove this

I assume this doesn't pass because it reverts with a generic EVM error?

test/Account.t.sol
```diff
@@ -498,6 +510,458 @@ contract AccountTest is BaseTest {
...
+    function testPayWithFiveCorruptedFieldOffsetsOfIntent() public {
+        bool success;
+        bytes memory returnData;
+
+        console.log("Test 1: Main Intent struct offset corruption");
+        bytes memory maliciousCalldata = _createIntentOnMainnet();
+        assembly {
+            let dataPtr := add(maliciousCalldata, 0x20) // Skip bytes length prefix
+            // CORRUPT MAIN OFFSET (Bytes 0-31) - Points to Intent struct start
+            mstore(dataPtr, 0x10000000000000000) // 2^64 (strictly greater than 2^64-1)
```
howydev, line 522: I'm not sure if works

If dataPtr is an encoded intent, wouldn't `add(maliciousCalldata, 0x20)` point to `intent.eoa`?

test/Account.t.sol
```diff
@@ -498,6 +510,458 @@ contract AccountTest is BaseTest {
...
+        assembly {
+            let dataPtr := add(maliciousCalldata, 0x20) // Skip bytes length prefix
+            let intentPtr := add(dataPtr, 0x20) // Points to start of Intent struct
+            // executionData offset (bytes 64-95 relative to start, or 32-63 in Intent struct)
+            mstore(add(intentPtr, 32), 0x10000000000000001) // 2^64 + 1
+        }
+        // assertEq(oc.execute(maliciousCalldata), bytes4(keccak256("VerifiedCallError()")));
+        (success, returnData) =
+            address(oc).call(abi.encodeWithSignature("execute(bytes)", maliciousCalldata));
+        assertEq(success, false);
```
howydev, line 539: could you think of a better check here instead of a generic execution failure? im a little worried that some of these tests are failing not because of corrupted calldata, so the test is inaccurate

test/Account.t.sol
```diff
@@ -498,6 +510,458 @@ contract AccountTest is BaseTest {
         assertEq(keysCount137, 2, "Keys should be added on chain 137");
     }
 
+    function testPayWithFiveCorruptedFieldOffsetsOfIntent() public {
+        bool success;
+        bytes memory returnData;
+
+        console.log("Test 1: Main Intent struct offset corruption");
```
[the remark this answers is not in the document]
smitrajput: fixed

test/Account.t.sol
smitrajput: removed.

yes, this is the EvmError: Revert error we get on corrupting every dynamic field offset <br/>
Screenshot 2025-09-03 at 4 45 57 PM

test/Account.t.sol
smitrajput: tried a bunch of approaches to catch the generic `EvmError: Revert` that we're getting on corrupting all 7 dynamic field offsets <br/>
Screenshot 2025-09-03 at 4 46 42 PM


but didn't find anything significant as that error returns `0x`, so further try/catching or error logging doesn't help. 

Also note that these tests are indeed failing because of offset corruption, as I [corrupted all the 13 other static fields](https://github.com/ithacaxyz/account/pull/312/files#diff-ff27a0826a73dea2fc4faf0dc2536085954313b31363b75b6db3a4a66d71ad99R543-R669) and we're seeing defined errors being thrown and caught for all of them.

smitrajput:
thanks a lot for the review @howydev 

I double checked to see if the offset corruption is correct, and we are indeed corrupting the exact offsets and we're seeing the generic `EvmError: Revert` uniformly for all the 7 dynamic offsets <br />
Screenshot 2025-09-03 at 4 46 42 PM

To help with the calldata corruption visualisation, plz check out this [calldata layout](https://github.com/ithacaxyz/account/pull/312/files#diff-ff27a0826a73dea2fc4faf0dc2536085954313b31363b75b6db3a4a66d71ad99R513-R538) in the tests, so we're sure we're corrupting the correct offsets.

To further prove that we're not getting those generic reverts bcoz of any adjacent fields' corruption, I corrupted all the remaining 13 static fields and we're throwing and catching well-defined errors for all of them. Plz find those tests [here](https://github.com/ithacaxyz/account/pull/312/files#diff-ff27a0826a73dea2fc4faf0dc2536085954313b31363b75b6db3a4a66d71ad99R543-R669).

This proves that the execution call fails differently for dynamic offsets corruption and static fields corruption.

Though you do have a valid concern that the generic `EvmError: Revert` might pop up for various other reasons, by observing those all other reasons (in this case the adjacent static fields' corruption) to be failing for other errors than this, we can safely say that all the uniform 7 `EvmError: Revert`s are indeed thrown by the 7 dynamic field offsets corruption.

howydev:
hey, apologies for the delay, didnt see that you replied

will allocate some time later this week for a review

howydev:
hey @smitrajput - really appreciate the time taken for this review, but we're likely going to switch to formats that remove all calldata offsets, using packed mode encoding instead of encoding a struct

By removing offsets, we kill 2 birds with a single stone - reduce the cost of operations, and no more security issues from custom abi encodings

smitrajput:
that's a cool idea @howydev, curious to have a look at the updated implementation and see if I can reuse these tests (at least the static fields corruption part and add the missing dynamic data corruption part too), to check the new and different security implications of the packed encoding

howydev:
changes are ready: https://github.com/ithacaxyz/account/pull/365

instead of having a struct that comes with offsets, we have a custom encoding for our `bytes` instead. without the offsets it's easier to reason about the security and we save 12k in gas when we do this, so we'll likely opt for some version of this instead of the original struct

would love your eyes on this if you have spare time, and if you think of any tests we should add feel free to propose!

smitrajput:
sound!

taking a look asap

PR #321, opened 2025-08-29
Add a direct function to execute pre calls

legion2002:


jxom (approved):

src/Orchestrator.sol
```diff
@@ -170,6 +170,42 @@ contract Orchestrator is
         TokenTransferLib.safeTransfer(token, recipient, amount);
     }
 
+    /// @dev DEPRECATION WARNING: This function will be deprecated in the future.
+    /// Allows pre calls to be executed individually, for counterfactual signatures.
+    function executePreCalls(bytes[] calldata encodedPreCalls) public virtual {
+        for (uint256 j; j < encodedPreCalls.length; ++j) {
+            // TODO: Add calldata bounds checks here.
+            SignedCall calldata p = _extractPreCall(encodedPreCalls[j]);
+            address eoa = Math.coalesce(p.eoa, msg.sender);
```
howydev, line 179: i think `msg.sender` is the relayer, right? so this might not work

we have `address(0)` defaulting to the account, will need to remove that optimization in this case

howydev (changes requested):
2 small fixes and it'll be good to go

src/Orchestrator.sol
howydev: coalescing with msg.sender doesnt rly achieves anything in that case right?

legion2002: It does, because you can still use preCalls signed with address(0), if you can call `orchestrator.executePreCalls()`, through the account.

So you don't have to get the user to sign a new precall if they have already signed one with eoa 0. You can just include the call to `orchestrator.executePreCalls()` in some other execution bundler that the account is already signing.

legion2002: Yeah good point, I updated the PR with the additional address field.
Think this is fine, because you can also do the same thing in the intent flow by just placing the precall inside a random eoa's intent.

howydev (approved):

PR #324, opened 2025-08-29
Remove `name` and `version` from 1271Sign domain

legion2002:
This is done to address a request from porto-sdk, which allows latency gains, since they don't have to dynamically fetch the name and the version of the account for each signature request.

Since this rehashing scheme is only meant to protect accounts where the same key is authorized on multiple accounts, just rehashing with the account address is enough for security.

Note: This breaks 5672 compatibilty, since the domain returned by the `eip712domain()` function does not match the domain used for the rehashing.

The domain separator used for the rehashing is - "EIP712Domain(address verifyingContract)"

Everything else remains the same.

### DIFF
--- a/src/IthacaAccount.sol
+++ b/src/IthacaAccount.sol
@@ -768,6 +768,6 @@ contract IthacaAccount is IIthacaAccount, EIP712, GuardedExecutor {
         returns (string memory name, string memory version)
     {
         name = "IthacaAccount";
-        version = "0.5.3";
+        version = "0.5.4";
     }
 }


src/IthacaAccount.sol
```diff
@@ -225,11 +225,37 @@ contract IthacaAccount is IIthacaAccount, EIP712, GuardedExecutor {
...
+            // Domain Typehash: `keccak256("EIP712Domain(address verifyingContract)")`
+            mstore(0x00, 0x035aff83d86937d35b32e04f0ddc6ff469290eef2f1b692d8a815c89404d4749)
+            mstore(0x20, address())
+            // Compute the digest.
+            mstore(0x20, keccak256(0x00, 0x40)) // Store the domain separator.
+            mstore(0x00, 0x1901) // Store "\x19\x01".
+            mstore(0x40, structHash) // Store the struct hash.
+            digest := keccak256(0x1e, 0x42)
+            mstore(0x40, m) // Restore the free memory pointer.
+            mstore(0x60, 0) // Restore the zero pointer.
```
howydev, line 247: can remove this

howydev (changes requested):
some changes needed

howydev (approved):

PR #325, opened 2025-08-29
feat: add git hooks

howydev:
add git hooks for better devX of running benchmarks

we use husky to automatically run pre-commit commands

changed benchmarks to use deterministically generated keys also, since --isolate causes gas to fluctuate depending on the number of 0 and non-0 bytes


src/IthacaAccount.sol
```diff
@@ -232,7 +232,6 @@ contract IthacaAccount is IIthacaAccount, EIP712, GuardedExecutor {
         virtual
         returns (bytes32 digest)
     {
-        /// @solidity memory-safe-assembly
         assembly ("memory-safe") {
```
howydev, line 235: this was throwing a solidity warning for doubling up on `memory-safe` tags

howydev:
yup like that idea, will grab it in a next one

PR #327, opened 2025-08-30
feat: fix and simplify multichain design

howydev:
`Intent.isMultichain` is used to switch between normal validation vs merkle root validation, whereas `Intent.nonce << 240 == 0xc1d0` is used to compute the digest without a chainId

This is potentially dangerous today. When both `Intent.isMultichain` and `Intent.nonce << 240 == 0xc1d0` is set, cross execution of intents is possible. When a user signs over a set of multichain intents containing an ethereum intent and a base intent, the base intent can be executed on Ethereum and vice versa

This PR contains a fix + a simplification, reducing gas by ~600 across all txs

Changelog:

1. Fix and simplify multichain - remove the `isMultichain` property and introduces a separate merkle verification nonce prefix `0x6D76`. An intent now can only be executed on a single chain with the normal verification mode, executed on pre-set number of chains with merkle verification, or replayable on all chains with multichain
-> Example use case: Merkle verification is used for cross chain execution (bridge + swap) and multichain is for account upgrades or ownership rotation (to apply allow your wallet to apply changes on chains that you will deploy on in the future)

2.  Remove redundant "is multichain" properties from both `computeDigest`s since the domain already contains chainId

3. Replace all usages of `Orchestrator.Intent` with `ICommon.Intent` to be consistent


.github/workflows/version-check.yaml
```diff
@@ -71,7 +71,7 @@ jobs:
           
           # Commit changes (only Solidity files, not package.json)
           git add src/*.sol
-          git commit -m "chore: bump contract versions due to bytecode changes - Contracts updated: $CONTRACTS_TO_BUMP"
+          git commit -m "chore: bump contract versions due to bytecode changes - Contracts updated: $CONTRACTS_TO_BUMP" --no-verify
```
howydev, line 74: skip husky pre hook

test/Orchestrator.t.sol
```diff
@@ -1482,6 +1479,7 @@ contract OrchestratorTest is BaseTest {
         vm.prank(t.gasWallet);
         t.errs = oc.execute(t.encodedIntents);
         assertEq(uint256(bytes32(t.errs[0])), 0);
+
```
howydev, line 1482:
```suggestion
```

PR #338, opened 2025-09-03
feat: self 7702 sigs shouldn't use a wrapper

howydev:


legion2002:
@howydev Can you add a changelog, and a PR description for what relay/sdk need to change once this PR is merged.

We should provide good docs for this, because I think it will change some stuff for them? Mainly I think they don't have to do rehashing for self EOA sigs now right?

legion2002 (approved):

howydev:
@legion2002 this contains a fix for husky so im going to merge it

PR #340, opened 2025-09-03
feat: allow funders to support older ocs

howydev:
# SimpleFunder: Support multiple orchestrators

  SimpleFunder now supports multiple orchestrators instead of a single immutable
  orchestrator. The `ORCHESTRATOR` immutable field has been replaced with an
  `orchestrators` mapping that can be managed by the contract owner.

  Deployment scripts updated to configure orchestrators after SimpleFunder deployment
   using the new `supported_orchestrators` config field.
   
   
   @kuyziss @jenpaff @laibe There is an additional config variable in the FundSigners script, which you have to set with all orchestrators, including the latest deployments.
   This step has to be completed before running anything interop related.

### DIFF
--- a/script/DeployAll.s.sol
+++ b/script/DeployAll.s.sol
@@ -26,18 +26,10 @@ contract DeployAllScript is Script {
         accountProxy = LibEIP7702.deployProxy(accountImplementation, address(0));
         simulator = address(new Simulator());
 
-        address[] memory orchestrators = vm.envOr("ORCHESTRATORS", ",", new address[](0));
-        address[] memory newOrchestrators = new address[](orchestrators.length + 1);
-        for (uint256 i; i < orchestrators.length; i++) {
-            newOrchestrators[i] = orchestrators[i];
-        }
-        newOrchestrators[orchestrators.length] = orchestrator;
-
-        funder = address(
-            new SimpleFunder(
-                vm.envAddress("FUNDER"), newOrchestrators, vm.envAddress("FUNDER_OWNER")
-            )
-        );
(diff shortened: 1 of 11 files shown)


script/DeployAll.s.sol
```diff
@@ -25,9 +25,17 @@ contract DeployAllScript is Script {
         accountImplementation = address(new IthacaAccount(address(orchestrator)));
         accountProxy = LibEIP7702.deployProxy(accountImplementation, address(0));
         simulator = address(new Simulator());
+
+        address[] memory orchestrators = vm.envOr("ORCHESTRATORS", ",", new address[](0));
```
legion2002, line 29: We don't use this DeployAll script for infra.
We should update this in the DeployMain.s.sol script in the deploy folder.

src/SimpleFunder.sol
```diff
@@ -43,10 +42,12 @@ contract SimpleFunder is EIP712, Ownable, IFunder {
     // Constructor
     ////////////////////////////////////////////////////////////////////////
 
-    constructor(address _funder, address _orchestrator, address _owner) {
+    constructor(address _funder, address[] memory _orchestrators, address _owner) {
```
legion2002, line 45: Think we should remove this from the constructor, otherwise it will give new simple funder addresses for all chains. 

We can just call the` setOrchestrator` function in the simple funder script in the deploy folder, where we set all the gas wallets.

We can set all old orchestrators there.

howydev: would the bytecode change mean we're going to have new simplefunder addresses?

PR #341, opened 2025-09-03
feat: add callSansTo support to ERC7821 execute

legion2002:


src/IthacaAccount.sol
```diff
@@ -485,6 +485,38 @@ contract IthacaAccount is IIthacaAccount, EIP712, GuardedExecutor {
...
+        returns (bytes32 result)
+    {
+        // If to is 0, it will be replaced with address(this)
+        assembly ("memory-safe") {
+            let t := shr(96, shl(96, to))
+            to := or(mul(address(), iszero(t)), t)
+        }
+
+        bytes32[] memory a = EfficientHashLib.malloc(calls.length);
+        for (uint256 i; i < calls.length; ++i) {
```
howydev, line 501: This means that someone can take a valid `CallSansTo` execution bundle and submit with the `Call` mode

Would prefer if we hash this a little differently to prevent that, its good design to only allow 1 possible way things can be executed with the same sig

test/Account.t.sol
```diff
@@ -68,6 +68,125 @@ contract AccountTest is BaseTest {
...
+            t.opData = abi.encodePacked(t.nonce, wrongSignature);
+            t.executionData = abi.encode(t.to, t.calls, t.opData);
+            vm.expectRevert(bytes4(keccak256("Unauthorized()")));
+            d.d.execute(_ERC7821_BATCH_SANS_TO_EXECUTION_MODE, t.executionData);
+            return;
+        }
+
+        d.d.execute(_ERC7821_BATCH_SANS_TO_EXECUTION_MODE, t.executionData);
+
+        assertEq(targetFunctionPayloads.length, t.n);
```
howydev, line 118: redundant test

howydev:
its not clear to me how this is flowing thru guarded executor, could you add a session key spend limit test with this new mode?

src/IthacaAccount.sol
legion2002: Yeah good point, but does it matter if they are used interchangeably?
the nonce already does replay protection. 

But yes maybe we should hash it separately, just to prevent footguns in the future

Vectorized:
Wait, let me marinate first. https://github.com/Vectorized/solady/pull/1485#issuecomment-3259620575

Imo, a native calldata compression like that in `cdCompress` will make this optimization automatically.

src/IthacaAccount.sol
howydev: nothing obvious, but yep best to not have hanging things like this, there are attacks that could happen when you combine a bunch of hanging things

PR #350, opened 2025-09-06
feat: improve benchmarks

howydev:
Improving benchmarks:
1. We weren't compiling porto's account in `via-ir`. The account used to be more yul but we dialed back into solidity, so we're seeing a lot of gains from this mode of compilation now (~1.8k gas reduction)
2. Cost of payment in any ERC20 token was ~17k higher across the board because the relayer did not have erc20 tokens (note: relayer is set to recipient in the pimlico paymaster case, so adding tokens to pimlico paymaster would not further reduce $)


gakonst:
If you'd be down - let's also add benches for direct account calls? Expecting smaller diff there but good to know

PR #365, opened 2025-09-14
feat: remove intent struct

howydev:
Instead of using an intent struct, we use a efficient custom encoding instead. This reduces gas by ~12k for single intents and ~10k for intents in a batch

Gas savings come from:
1. Smaller intent size (by ~5-10 words). This results in less memory usage, less memory expansion, and less intrinsic calldata cost paid
2. Removing custom offset checks `LibBytes.checkInCalldata` (~2-3k gas alone)
3. Not requiring just-in-time checks on struct offsets when accessing any struct field (solc adds this under the hood)
4. Optimizing the args for `_pay` to only take in what we currently use today instead of the full intent
5. Removing fields from the EIP-712 digest - 2 less `keccak` operations

The new format is `abi.encodePacked`, but with some nuances:
1. All static fields go before all dynamic fields, so all static fields have known offsets
2. The static fields that go into the EIP712 digest are ordered continuously
3. The static fields that are in the EIP712 digest are padded to 32 bytes, then with the previous property this allows us to do a single `calldatacopy` for digest calculations
(The relayer has the freedom to use dirty bytes on these fields, but then the digest would compute to something else and this would fail verification, so there's no incentive to do so)
4. Dynamic fields are prefixed with `uint256 field.length` to allow us to use them as a `bytes calldata` directly
5. We order the dynamic fields in the order that they will be used in the orchestrator flow

We rely on the helper function `_getNextBytes(ptr)` for efficient traversal through the dynamic fields. The helper function assumes that the pointer already points to the start of a dynamic byte range. It parses that out, then updates the CalldataPointer in memory to point to the start of the next dynamic byte range


src/Orchestrator.sol
```diff
@@ -236,11 +244,16 @@ contract Orchestrator is IOrchestrator, EIP712, CallContextChecker, ReentrancyGu
...
     /// @return gasUsed The amount of gas used by the execution. (Only returned if `isStateOverride` is true)
-    function simulateExecute(
-        bool isStateOverride,
-        uint256 combinedGasOverride,
-        bytes calldata encodedIntent
-    ) external payable returns (uint256) {
+    function simulateExecute(bytes calldata encodedIntent) external payable returns (uint256) {
+        bool isStateOverride;
+        uint256 combinedGasOverride;
+        assembly ("memory-safe") {
```
howydev, line 250: note to reviewer - we need to do this b/c adding another arg changes the calldata layout, making encodedIntent have an offset that is not `0x40`

src/Orchestrator.sol
```diff
@@ -20,12 +20,13 @@ import {ICommon} from "./interfaces/ICommon.sol";
...
 import {ISettler} from "./interfaces/ISettler.sol";
 import {MerkleProofLib} from "solady/utils/MerkleProofLib.sol";
+import {IntentHelpers} from "./libraries/IntentHelpers.sol";
 
 /// @title Orchestrator
 /// @notice Enables atomic verification, gas compensation and execution across eoas.
 /// @dev
 /// The Orchestrator allows relayers to submit payloads on one or more eoas,
-/// and get compensated for the gas spent in an atomic transaction.
+/// and get compensated for the gas spenst in an atomic transaction.
```
howydev, line 29:
```suggestion
/// and get compensated for the gas spent in an atomic transaction.
```

test/Orchestrator.t.sol
```diff
@@ -88,11 +89,9 @@ contract OrchestratorTest is BaseTest {
         u.paymentAmount = 0.1 ether;
         u.paymentMaxAmount = 0.5 ether;
         u.combinedGas = 10000000;
-        u.signature = "";
-
         u.signature = _sig(alice, u);
 
-        assertEq(oc.execute(abi.encode(u)), bytes4(keccak256("PaymentError()")));
+        assertEq(oc.execute(encodeIntent(u)), bytes4(keccak256("Unauthorized()")));
```
howydev, line 94: found it funny that a `testUnauthorized` was erroring with `PaymentError`

test/utils/mocks/MockPayerWithSignatureOptimized.sol
```diff
@@ -6,6 +6,7 @@ import {Ownable} from "solady/auth/Ownable.sol";
 import {ECDSA} from "solady/utils/ECDSA.sol";
 import {ICommon} from "../../../src/interfaces/ICommon.sol";
 import {IOrchestrator} from "../../../src/interfaces/IOrchestrator.sol";
+
```
howydev, line 9:
```suggestion
```

src/IthacaAccount.sol
```diff
@@ -753,6 +738,6 @@ contract IthacaAccount is IIthacaAccount, EIP712, GuardedExecutor {
         returns (string memory name, string memory version)
     {
         name = "IthacaAccount";
-        version = "0.5.7";
+        version = "0.5.8";
```
howydev, line 741:
```suggestion
        version = "0.5.7";
```

src/SimpleFunder.sol
```diff
@@ -56,7 +56,7 @@ contract SimpleFunder is EIP712, Ownable, IFunder {
         returns (string memory name, string memory version)
     {
         name = "SimpleFunder";
-        version = "0.1.6";
+        version = "0.1.7";
```
howydev, line 59:
```suggestion
        version = "0.1.6";
```

legion2002:
In general I really like the idea. 
It has some UX tradeoffs for integrators, but most integrators will be sophisticated like the relay, and it should be pretty easy for them to handle the complexity.

We can also provide pure functions in the library that will convert Intent structs into the custom encoded form, so integrators can just RPC call into this.

It does add some complexity, but I think it is justified by the gas savings we get from this. Also to note, we are getting rid of all the calldata offset check issues, which is actually a net security improvement even if we add explicity complexity to the decoding.

src/libraries/IntentHelpers.sol
```diff
@@ -0,0 +1,161 @@
...
+            funder := calldataload(data.offset)
+            sig.offset := add(data.offset, 0x80)
+            sig.length := calldataload(add(data.offset, 0x60))
+            // transfers array starts after sig, but needs to account for padding
+            let sigLenWords := mul(div(add(sig.length, 31), 32), 32)
+            transfers.length := calldataload(add(sig.offset, sigLenWords))
+            transfers.offset := add(add(sig.offset, sigLenWords), 0x20)
+        }
+
+        // update data offset to after the transfers array
```
howydev, line 155: note to self to remove

this was necessary before transitioning to using a memory ptr

src/Orchestrator.sol
```diff
@@ -224,7 +231,7 @@ contract Orchestrator is IOrchestrator, EIP712, CallContextChecker, ReentrancyGu
             // We reluctantly use regular Solidity to access `encodedIntents[i]`.
             // This generates an unnecessary check for `i < encodedIntents.length`, but helps
             // generate all the implicit calldata bound checks on `encodedIntents[i]`.
-            (, errs[i]) = _execute(encodedIntents[i], 0, _NORMAL_MODE_FLAG);
+            errs[i] = this.execute(encodedIntents[i]);
```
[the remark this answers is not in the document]
howydev: yup, its possible, this approach is less efficient

right now its 10k cheaper in a batch and 12k in a single intent, i think that 2k difference is coming from this

let's include that in a separate PR, it might be a medium-sized effort if we're going to go through with some of the other planned changes

PR #372, opened 2025-09-17
feat: add spend limit enabled flag

jxom:
This allows non super admin keys, to skip spending limit checks IF a super admin key authorizes the key in this way.

This is useful if you want to have a session key for some period of time which can do any actions on your account. 

This is also safer than authorizing/revoking a super admin key because this temporary unlimited spending key cannot authorize/revoke other keys.


legion2002 (dismissed):

gakonst (commented):
can we also test that the upgrade is safe? maybe etch the bytecode of previous version?

howydev (approved):

PR #376, opened 2025-09-18
chore: remove devtools dependency

legion2002:
The devtools repo takes too long to load, this PR removes it as a dependencies, and copy pastes the relevant code into the file.


gakonst:
> Note: devtools will still be installed atm, because we also install the account_v0.5.5 as a submodule. We will patch this separately in a different PR.

Think just fork at block & use existing account vs having to maintain a bytecode to etch.

legion2002:
> Think just fork at block & use existing account vs having to maintain a bytecode to etch.

Yup makes sense, will do in a separate PR.

PR #380, opened 2025-09-18
fix: prevent duplicate signatures from same owner in MultiSigSigner

avorylli:
### SECURITY FIX: Prevent multisig bypass through duplicate signatures

Fixes vulnerability where a single owner could provide multiple
signatures to bypass the multisig threshold requirement.

### VULNERABILITY:
- The original implementation allowed the same owner to provide multiple
  signatures in a single transaction, each counting towards the threshold
- This enabled a single owner to bypass the multisig protection by
  duplicating their own signature

### CHANGES:
- Added tracking of used owners to prevent duplicate signatures
- Each owner can now only contribute one signature per validation
- Maintains backward compatibility

### IMPACT:
- Prevents unauthorized multisig bypass attacks
- Ensures proper multisig threshold enforcement


avorylli:
@legion2002 seems to be you're right - I misunderstood the duplicate owner functionality. Closing this PR.

Issue #385, opened 2025-09-22
[FEATURE]: Benchmarks 2.0


howydev:
### Feature Summary

The current benchmarks accurately benchmark Porto vs 4337 accounts

The benchmarks are split into 2 categories - a single intent/UOs, and a batch of 100 intents/UOs. The goal was to have our benchmark in line with existing widely used benchmark numbers like alchemy's aa-benchmark, but also showcase real-world costs of using an AA solution

Some of the choices are not in line with the latter. For instance, we should use a non-zero nonce in the real-world benchmarks. Since we now use foundry's `snapshotGasLastCall`, this allows us to refactor + consolidate each benchmark into a single function for all accounts, which would make it much easier to work on

### Timeline

Improvement (short term)

### Security Considerations

n/a

howydev:
Stretch goal: Benchmark porto account in 7702 mode also, instead of 7702 -> proxy -> impl

howydev:
related: https://github.com/ithacaxyz/account/issues/335

PR #391, opened 2025-09-26
feat: update benchmarks

howydev:
1. Iterate on our benchmarks to use "maximum cost" and "average cost"
a. Average cost - benchmarks 2nd+ intent/UO from an account, batched in a batch of 10
b. Maximum cost - benchmarks 1st intent/UO, unbatched

2. Update the optimized app sponsor cost for ithaca accounts


legion2002 (commented):
I am okay with merging this, but don't think Maximum cost benchmarks have much value.
User will only encounter these costs once in their lifetime, so we wouldn't optimize this

We should just do average cost benchmarks with sensible nonce and other storage slot options.

legion2002 (approved):

howydev:
semi-related issue: https://github.com/ithacaxyz/account/issues/385

PR #393, opened 2025-09-29
feat(v1.0.0): merkle

legion2002:
Applies PR #388 to v1.0.0 account.

With an additional breaking change, which removes the orchestrator's redundant `verifyMerkleSig` function, and unifies all flows through the account's native merkle sig verification.

Note: large diff is because of new foundry fmt changes, the main changes are in the IthacaAccount, Orchestrator, Account.t.sol, and Orchestrator.t.sol files

### DIFF
--- a/src/Orchestrator.sol
+++ b/src/Orchestrator.sol
@@ -859,7 +859,7 @@ contract Orchestrator is
         returns (string memory name, string memory version)
     {
         name = "Orchestrator";
-        version = "0.5.5";
+        version = "0.5.6";
     }
 
     ////////////////////////////////////////////////////////////////////////
--- a/src/SimpleFunder.sol
+++ b/src/SimpleFunder.sol
@@ -58,7 +58,7 @@ contract SimpleFunder is EIP712, Ownable, IFunder {
         returns (string memory name, string memory version)
     {
         name = "SimpleFunder";
-        version = "0.1.8";
+        version = "0.1.9";
     }
 
     ////////////////////////////////////////////////////////////////////////


src/IthacaAccount.sol
```diff
@@ -478,38 +479,80 @@ contract IthacaAccount is IIthacaAccount, EIP712, GuardedExecutor {
...
+
+    /// @dev Verifies the merkle sig
+    /// - Note: Each leaf of the merkle tree should be a standard digest.
+    /// - The signature for using merkle verification is encoded as:
+    /// - bytes signature = abi.encodePacked(bytes32[] proof, bytes32 root, bytes rootSig)
+    function _verifyMerkleSig(bytes32 digest, bytes calldata signature)
+        internal
+        view
+        returns (bool isValid, bytes32 keyHash)
+    {
```
howydev, line 497: continuing convo from this: https://github.com/ithacaxyz/account/pull/388/files#r2387828490

I think the outer one is entirely ignored, so theres no security issue

`return _verifyMerkleSig(digest, signature)` -> this overrides the `keyHash = ...` parsed in the top level call frame

On prehash - IIUC there should only be 1 `prehash` that's set, in either the outer or inner wrapper. There's a difference between the 2 since the digest has to be in the merkle tree, so which one we want to set depends whether the merkle leaf is `sha`-ed or not. I don't know which one would be `sha`ed in practice, but the other one would probably be redundant

howydev: Something that I just thought to check was whether we'll run into issues around [second preimage attacks](https://rareskills.io/post/merkle-tree-second-preimage-attack) 

I don't see anything thats obvious so far, bringing it up just in case im missing something

PR #402, opened 2025-10-15
feat: add multicall to simulator

howydev:


src/Simulator.sol
```diff
@@ -125,6 +126,90 @@ contract Simulator {
...
+                // If any pre-call failed, we return gasUsed = 0 and the error data from that call
+                if (!results[i].success) {
+                    return (0, results[i].returnData);
+                }
+            }
+        }
+        IMulticall3.Result memory lastResult = results[results.length - 1];
+        lastReturnData = lastResult.returnData;
+
+        // Parse gasUsed from the orchestrator's return data
```
joshieDo: this is only returning the `gasUsed` by the intent right? would be interested in the overal gasUsed of the whole multicall including the intent and precalls

joshieDo: actually both. used by intent (so we can cap its execution) and by intent+precalls so we can price it

joshieDo (approved):

